PR #24, opened 2022-10-22
Implemented change Dashboard Button

Easylemon24:
Added a Button that only shows if the User is admin and then leads the User to the Admin dashboard.
Added a Button to the Admin dashboard that redirects to the Userdashboard


Easylemon24:
Added also some fixes that where ccaused by typos.
Added a little display for a credit system.
Added that the Tickets of an user are displayed at the Users Dashboard

Easylemon24:
Fixed that the show credits display was in a wrong sub div.
Added a parameter to filter closed tickets.

PR #29, opened 2022-11-06
Fixed Products drop-down not closing

ItsLeon15:


ItsLeon15:
Added `home.blade.php` as well. The file `home.css` is bootstrap which is meant for the table. CSS isn't my best so someone can try and fix the resizing of the page.

CorwinDev:
If you try: npm run build:default the css will be build, and compiled.
So dont use more css files :)

PR #38, opened 2022-11-08
Changed ticket style

ItsLeon15:


CorwinDev:
GG, next step is removing the ugly ticket view

PR #40, opened 2022-11-08
Added option to change password

ItsLeon15:
Removed ``<script src="https://cdn.tailwindcss.com"></script>`` so the change-password page will look off. Not sure why ``public/default/assets/app.ad4b13dd.css`` has changed but you can remove it if it doesn't break anything. 

Redirects to `/home` on success. Can be changed if you want.


CorwinDev:
Please move the password.blade.php to the auth folder, as it isn't a layout file

PR #77, opened 2023-03-01
pt-translation

fpaiva36834:
Here's a translation to portuguese


fpaiva36834:
If there's any other translation in need just tell me :D

PR #81, opened 2023-03-11
Updated HU translation

SprayerGaming:
Updated HU.json translation due to some new english words  phrases was added.


SprayerGaming:
+ Updated SK.json - same reason

SprayerGaming:
Fixed un-translated / badly translated words

CorwinDev:
Could you only add the commits

PR #91, opened 2023-04-07
Fixed czech translations

ecocraft:
Added a dash in e-mail and fixed some goofy translations.


ecocraft:
Found some more goofy translations

PR #94, opened 2023-04-15
That it has been possible to deactivate the white mode or do it as the main one. And add a new theme.

OverNodes-net:
That you can deactivate the white mode or do it as the main one. And add a new theme.


CorwinDev:
Sorry but why is this a pull request?

OverNodes-net:
I was in the wrong place.

PR #111, opened 2023-05-30
Throw exception instead of returning true

GJSBRT:


CorwinDev:
Some themes doesn’t have a config (yet) so we don’t want to throw a error if there isn’t any

GJSBRT:
I don't agree with you. Themes should be updated anyway, because of new functionality in newer versions

PR #119, opened 2023-06-15
Update HU.json

SprayerGaming:
Recently updated HU translation to match the new English sentences were added in the past.


SprayerGaming:
+ Recently finished update of SK.json and fixed wrong translation of password.php file in dir SK/password.sphp

PR #133, opened 2023-08-17
add social registration

NoPantsMcDance:
When a social user logs it, it creates an account for them using their email. 

If the user goes to a page that requires a password auth, it bypasses the check. (Could force them to relogin with the social, seemed like a bit much for my needs)

Added migration to have a "is_social_user" flag set during registration.


Turkidev:
Make it sends an email to set there password if they want to

PR #135, opened 2023-08-21
Fixing proxmox not showing the saved vmid

Turkidev:


GJSBRT:
Why all the unnecessary formatting changes? Please only change actual code, it makes reviewing way easier

app/Extensions/Servers/Proxmox/Proxmox.php
```diff
@@ -759,21 +771,25 @@ public function terminateServer($user, $parmas, $order, $product, $configurableO
...
+            throw new Exception('Unable to terminate server');
         return true;
     }
 
 
     public function getCustomPages($user, $parmas, $order, $product, $configurableOptions)
     {
         $vmType = $parmas['type'];
-        $vmid = $parmas['config']['vmid'];
+        $vmid = ExtensionHelper::getOrderProductConfig('vmid', $product->id);
```
GJSBRT, line 783: Add a check if a value is returned

GJSBRT (changes requested):

Turkidev:
sorry its auto formatting in vscode

app/Extensions/Servers/Proxmox/Proxmox.php
Turkidev: check if its null?

GJSBRT: Yes, In your helper function you return if there is no value, which should also be a `return null;` btw.

Turkidev: if ($vmid == null) return;?

PR #138, opened 2023-08-22
Out of stock UPDATED

Turkidev:
when the product is out of stock show it in the store


CorwinDev:
Please only edit the neccesary code, not anything else

Turkidev:
see this now

PR #142, opened 2023-08-23
Show hostname and ip for proxmox

Turkidev:
this pr will add ip and hostname if its proxmox server


CorwinDev:
So you want users to open that file to edit the ip?
And why not adding it to the proxmox view?

Turkidev:
the admin after setting the ip he can add it to the panel
here is better

PR #149, opened 2023-09-14
Some UI changes, new paymenter version is available for download notify

mscodepl:
image


mscodepl:
Only for admin, shows throughout the panel, if not now clicked it disappears for 12 hours

app/Console/Kernel.php
mscodepl, line 1: use App\nCheckUpdates; has been removed in the new commit :P

PR #169, opened 2023-10-18
Improvements

ProgyNL:
Improving dutch language


CorwinDev:
Why removing so much language strings from en.json?

ProgyNL:
> Why removing so much language strings from en.json?
First time using pull request. Didn't remove anything i just updated it.

PR #281, opened 2023-12-03
[Spanish] Updated es.json with the latest strings from master.

mendozal:
I created a development environment and I ran:
npx laravel-language-extractor --theme default 

And:

- Translated the new added strings from the master branch. 
- Made a few fixes.

To the es.json file.


mendozal:
I'm creating another pull request with these fixes + solution for settings tabs translation

PR #329, opened 2024-01-07
Remove check to not return free product invoices. 

NoPantsMcDance:
Already checks if the total is 0 in the cronjob, this check is redundant and will cause a loop if you try to invoice free products as it will never return an already existing invoice for a free product


ProgyNL:
If im not wrong, the removed part (what you did) belongs to invoice_items and the other part from cronjob is controlling the invoice itself.

The removed part sends an 'signal' to cronjob that there is no charge so its need to be marked as paid when orderProduct sends 0, otherwise it would be pending and the customer received an email to pay $0 i guess?

PR #366, opened 2024-02-04
Color errors red and no longer make use of the `$DEFAULT_URL` variable.

prplwtf:
Currently there are two variables for defining the release URL for the `update.sh` script assigned the following values:

`$URL` https://github.com/paymenter/paymenter/releases/latest/download/paymenter.tar.gz
`$DEFAULT_URL` https://github.com/paymenter/paymenter/releases/latest/download/paymenter.tar.gz

When `$URL` is empty, the value of `$DEFAULT_URL` gets assigned to `$URL`, but this never happens unless made completely empty by editing the file rendering a specific "if" statement completely useless.

The changes made in this pull request will instead throw an error when `$URL` is empty, which only happens if a user would edit the file and remove the value of `$URL`. I've also colored errors red because why not.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -60,7 +60,7 @@ if [ -t 0 ]; then
         fi 
     fi
     
-    if [ -z $INSTALL ]; then
+    if [ -z "$INSTALL" ]; then
         read -p "Are you sure you want to run the upgrade process for your Panel? [y/N]: " -r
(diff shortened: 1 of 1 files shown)


prplwtf:
Somehow this includes my previous pull request for some reason. It should be fine because I merged the changes from previous commits to this file as well, but it's still strange.

PR #370, opened 2024-02-07
add fr lang

turnbass-hyzux:


CorwinDev:
Please dont translate "Paymenter" to "payeurs" just keep it Paymenter

PR #375, opened 2024-02-18
Local development environment.

ChecksumDev:
Working on Arch Linux, need testing on other systems before merge, but I'm fairly positive it will work.


ChecksumDev:
Not sure why this pushed here, ignore.

PR #402, opened 2024-03-23
Fixed Timezone Issue

Sarthak-07:


CorwinDev:
Please read it from settings

Sarthak-07:
My settings had Asia/Kolkata since starting but it keeps saying UTC, well adding this fixed the issue for me.....

PR #408, opened 2024-03-30
Update Danish Translations

PhillipEis:
I've updated alot of the translations in them, both missing and miss worded


Justman100:
@CorwinDev

PR #414, opened 2024-04-05
Create debug command

Poseidon281:
I have made a very simple debug command that can be triggered by doing 

cd /var/www/paymenter && php artisan debug

This command simply enables or disables paymenters debugmode


CorwinDev:
Hi,
I would not recommend enabling debug mode in production, and as all logs are available under storage/logs I don’t think this is necessary.

Debug mode is useful if you’re developing in paymenter (theme, contributions etc) and I hope developers can enable it themself ;)

PR #437, opened 2024-05-14
Added support images with s3 storage

piratetm:
✅This change will add support for distributed container management. 
✅Tested on k8s, on 5 machines.

Added to:
- Categories
- Products
- Logo

image
image


piratetm:
image

PR #448, opened 2024-06-01
feat: add admin credits api endpoint and permissions

VaibhavSys:


VaibhavSys:
Seems like no updates to 0.x versions now

63zee:
Corwin is not going to make any more changes to 0.x versions because the code sucks.... He only makes changes when it is important and/or fast, so he is working on V1

PR #454, opened 2024-06-10
Resolve issue with multiple IP addresses not displaying and IPv6 showing as 'N/A'.

ImAubrey:
Resolve issue with multiple IP addresses not displaying and IPv6 showing as 'N/A'.


CorwinDev:
Mind using blade syntax instead of raw php opening and closing?
`<?php` = `@php`
`?>` = `@endphp`

Same for echoing variables:

`<?php echo $ipv4_string; ?>` = `{{ $ipv4_string }}`
As it contains <br /> you might need to use `{!! $ipv4_string !!}`

PR #482, opened 2024-07-18
Add new function for Virtfusion

ImAubrey:


ImAubrey:
> Fix conflicts 🙈

I don't really think that the current version can be more powerful or even a downgrade.

I prefer to use Paymenter to manage the product attributes directly, for example, to control the Core or some details in the Admin panel. It can be a simple deployment process, I don't need to create a package again in Virtfusion. 

If someone prefers to use package manage in Virtfusion, empty the attributes, and products will follow settings from the original Virtfusion Package center defined

imgucciwz:
I understand your point, but you can create a simple package on VirtFusion and override the specs through configuration options if needed. Skipping a step isn't necessary since VirtFusion requires a package to create a VPS. Keep something similar like Virtualizor. Why ask uneccesary information and not keep it clean?

PR #526, opened 2024-10-30
fix: home and dashboard navigations

Sarthak-07:


CorwinDev:
Please don't hardcode routes, place them in the navigation class

PR #529, opened 2024-10-30
feat: turnstile theme and flex size

iamdarkle:
Adds widget size adjustment via API that takes up all available space (100% with a minimum of 300px), and also check the dark mode to adapt the widget theme.


iamdarkle:
It's present in the html tag, isn't it?

CorwinDev:
image
Nope it isn't, lets default to dark theme

PR #545, opened 2024-11-12
Fix department indexing inconsistency between admin and user dashboards

BILALMRN:
- Resolved an issue where the department was stored as an index in the admin dashboard and as a string in the user dashboard.
- Updated code to consistently handle the department using an integer index across both dashboards.
-  The department is now stored as the index in the database.


CorwinDev:
Lets store default as string and not as index

PR #550, opened 2024-11-26
feat: allow users to download invoices pdf

Sarthak-07:
Allows users to download invoices pdf and also adds invoice id in the pdf name.
image


CorwinDev:
Doesn't really look good in that place, can you try just in text on the right top?

Sarthak-07:
Done!
image

PR #582, opened 2025-02-12
Some tweaks for Tailwind v4

iamdarkle:
For extensions I used `@source "../../../extensions/**/*.blade.php";` should it point only to the folder or otherwise? 🤔


themes/default/css/app.css
```diff
@@ -31,5 +64,3 @@ color utility to any element that depends on these defaults.
 @utility text-wrap {
   text-wrap: wrap;
 }
-
-@import './easymde.css'
```
CorwinDev, line 35: Without EasyMDE css file ticket textarea's are not working

CorwinDev (changes requested):

PR #589, opened 2025-02-23
Added function for selecting invoice management

officialPlocki:
Added Sevdesk support (untested)

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/Paymenter/Paymenter/pull/589?shareId=a7eca54e-00db-4673-94cc-014c97b48c04).


officialPlocki:
I want to explain the reason behind this push: Due to EU regulations and laws in Germany/Austria/Swiss, we need a GoBD (german law) conform invoice system.
This means that it must be able to create non-deleteable (but cancelable) invoices, a strict protocol of actions and comply to many more requirements.

If this is not integrated, it is advised to be used by companies around here, without agreeing to get hefty fines.

This pull can contain some errors (i cant code in PHP), so I started it with a edit via copilot.
I love the idea of the panel and would like to use it for my future hosting, but cant do so without it.

So please take it as a "short" feature request, because i think it would also be nice if there would be a option for all users to change the billing system between internal, Sevdesk, Billbee and so on.
Or just have a integration ability for it that we can add a handler to it without bigger efforts.

I tested it and got a manifest error with vite and some other ones.

Would appreciate that feature 💯

CorwinDev:
I do not think this should be included as core code, you might want to release this as a extension on the marketplace or use privately on your own panel

PR #634, opened 2025-03-16
fix: convoy malformed url not working

wjy20030407:


CorwinDev:
With this code it’s adding a duplicate / to the url

wjy20030407:
> With this code it’s adding a duplicate / to the url

This change removes (not adds) the leading slash, fixing the current double slash issue in URLs like "api/application//users/1/generate-sso-token"

CorwinDev:
Oops 😬

PR #690, opened 2025-04-12
feat: Add invoice reminder logic for services expiring in x days

gps949:


CorwinDev:
I think this should be a different email template with a text saying: "This is a reminder that your invoice made on <x date> is still unpaid, please make sure to pay it to avoid any service interruptions."

gps949:
> I think this should be a different email template with a text saying: "This is a reminder that your invoice made on is still unpaid, please make sure to pay it to avoid any service interruptions."

I completely agree with you. I’ll submit a separate PR for this, so we can close this one.

PR #698, opened 2025-04-14
feat: Add invoice reminder logic for services expiring in x days

gps949:
Previously, there was a setting in the admin panel for unpaid invoice reminders, but there was no actual implementation in the CronJob to trigger such reminders.

This PR adds the missing logic:
	•	Introduces a new email template for invoice reminders
	•	Adds the corresponding Invoice Reminder event and listener
	•	Updates the CronJob to check services that are still in active status and have associated invoices in pending status. If the service is set to expire in X days (based on admin settings), it triggers the InvoiceReminder event to send a reminder email to the user

This ensures users are properly notified about unpaid invoices before their services expire.

### DIFF
--- a/app/Console/Commands/CronJob.php
+++ b/app/Console/Commands/CronJob.php
@@ -3,6 +3,7 @@
 namespace App\Console\Commands;
 
 use App\Events\Invoice\Created as InvoiceCreated;
+use App\Events\Invoice\Reminder as InvoiceReminder;
 use App\Jobs\Server\SuspendJob;
 use App\Jobs\Server\TerminateJob;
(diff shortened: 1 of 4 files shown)


CorwinDev:
Do we need a event for that? Think only sending notification would be enough

gps949:
> Do we need a event for that? Think only sending notification would be enough

Adding a new event might not be strictly necessary, but keeping the logic consistent with other events could help maintain internal consistency and scalability of the system. You’re free to decide the best approach. If adding an event feels too verbose, feel free to go with whatever works best for you.

Issue #662, opened 2025-04-06
Upgrade button goes after invoice generation


CpuSocket:
If you try to upgrade a server press a upgrade then press " Pay " after it generates the invoice the upgrade button disappears meaning you cant change what upgrade you want until you pay the other upgrade invoice.

Basically:

I want to upgrade so i press upgrade to next level plan
I press pay on upgrade and it generates invoice
Say i change my mind and want to upgrade to another level plan i cant because the button is gone for upgrade.

CpuSocket:
This could be whats meant to happen tbf as i think WH*** is the same but a nice popup like WHMCS does like 

WHMCS message:

" Upgrade/Downgrade not available

You cannot currently upgrade or downgrade this product because an upgrade or downgrade is already in progress.

To proceed, please first pay the outstanding invoice "

PR #704, opened 2025-04-15
(feat) Show notification when there is an invoice open

Scraayp:
Don't allow a user to upgrade his service when there is a invoice open.

### DIFF
--- a/app/Livewire/Services/Upgrade.php
+++ b/app/Livewire/Services/Upgrade.php
@@ -58,8 +58,15 @@ class Upgrade extends Component
 
     public function doUpgrade()
     {
+        $user = Auth::user();
         if (!$this->service->upgradable) {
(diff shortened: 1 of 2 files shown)


Scraayp:
Should probably add a setting for this, so that the end-user can choose it himself.

Scraayp:
Should probably add a setting for this, so that the end-user can change it himself.

Scraayp:
Solves: #662 .

CorwinDev:
#662 is a different issue

And why are the changes in app/Models/ServiceUpgrade.php neccesary?

CorwinDev:
In my opinion one time purchases shouldn't be upgradable as they are not a service/subscription

CorwinDev:
But there is something to say if people really want that, please make a new pull request with only that part included


And this pr doesn't fix #662

PR #720, opened 2025-04-21
Fix duplicate invoice email and missing item lines on admin invoice creation

### LINKED ISSUE #663: Invoice formatted price wrong
When generating a invoice for a user manually it generates the invoice for the correct amount and stripe reads it as the correct amount.
but in emails it will say £/$ 0 and also on Discord webhooks.

( Invoice page showing correct amount ) 
Image

Email showing $/£ 0

Image

Moutard3:
The problem was that the CreateEvent was fired before Filament inserted the related rows. I added a virtual property in the Invoice model, and added a dispatch to fire the event after Filament inserted the Item rows.

Also fixes the duplicate mail sending, as it was sent by both the event and the handleRecordCreation method (and it is better imo to keep the email sending logic in the event)

Fix #663

### DIFF
--- a/app/Admin/Resources/InvoiceResource/Pages/CreateInvoice.php
+++ b/app/Admin/Resources/InvoiceResource/Pages/CreateInvoice.php
@@ -3,7 +3,6 @@
 namespace App\Admin\Resources\InvoiceResource\Pages;
 
 use App\Admin\Resources\InvoiceResource;
-use App\Helpers\NotificationHelper;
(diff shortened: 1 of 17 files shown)


CorwinDev:
I really like this approach but maybe we can optimize it a bit more? Lets make a seperate event called `Finalized` and trigger that instead of the created event with sendmail tag, this allows for 3rd party extension to also listen do things after all the items are added

Also if you want, you can also update the Cart.php invoice creation and CronJob.php invoice creation

PR #722, opened 2025-04-23
feat: Support for hidden custom properties.

ronaldvaneede:
For a extension I am working on I need to store an ID from a third party system on the User model.
This makes it possible to create properties on the user (or any other model that supports custom properties) that can be used for those kind of use cases.
The property will be hidden from the Personal Details screen but is visible and editable in the admin interface.


CorwinDev:
You can already make properties hidden on the user model,

ronaldvaneede:
oh I was not aware of that, it does, never mind then 😄

PR #724, opened 2025-04-24
feat: Invoice numbers with formatting

ronaldvaneede:
Instead of using the Invoice model id this adds a 'number' attribute that uses the ID and adds a optional prefix and padding.
With this is is possible to have invoice numbers like 0001, 0002 etc, or 20250001, 2025002 etc or INV202504240001 etc.
The padding and prefix can be configured in a dedicated section in the system settings and currently supports 3 placeholders: {year}, {month} and {day}.

invoice1
invoice2
invoice3
invoice4


ronaldvaneede:
Good suggestions, I will work on that and update the PR soon.

ronaldvaneede:
This should do it, I put the generation in the creating static method in the Invoice model boot() method. 
Let me know if you prefer it in a Listener class for the creating event, then I will make that change.

ronaldvaneede:
I made some updates to implement the listener and some other improvements.
The settings are not necessarily required, if they are not defined the defaults are used with a padding of 1 and {number} as format. So essentially the same as the id.
I had the migration already but forgot to include it in the commit :)

ronaldvaneede:
Done 😄

PR #742, opened 2025-04-29
feat: Account created and acount updated events

ronaldvaneede:
Adds account created and account updated events that are dispatched after the user has been created and the properties have been set and right after the user has been updated with its properties.
This is in addition to the User created and User updated events which are dispatched before the properties are set.
This allows extensions to hook into account create and update events and have access to the properties on the User model which are not there with the User created and User updated events.


CorwinDev:
I honestly would add property events instead of hardcoded events for register

CorwinDev:
Whats your usecase if I may ask?

ronaldvaneede:
For the registration that will work if we dispatch the User/Created event after the response using the afterResponse() in the UserObserver. That works for me, then I can get the properties in the listener.
I have a bit of trouble with the update if we switch it around as proposed because the updated event is only dispatched when the User model is dirty. 
That only happens when the first_name, last_name or email are changed but not the other properties.

CorwinDev:
I see, lets make a seperate event for any properties that are updated: https://github.com/Paymenter/Paymenter/blob/master/app/Livewire/ComponentWithProperties.php#L81 (this file is used for any page that uses account/profile properties so we don't have to trigger events seperatly)

We can drop the Register event as User/Created does the same

ronaldvaneede:
I made some changes and added an event at the bottom of the updateProperties() method in the ComponentWithProperties to dispatch an event with the user and all it's properties when a user registers or updates his profile and also added an observer to the Property model to make it possible to respond on individual property changes (for example trough the admin ui) if needed.

CorwinDev:
LGTM, lets remove app/Events/Properties/Updated.php debug code :D

ronaldvaneede:
🚀

PR #744, opened 2025-04-30
Update SECURITY.md

aadi2206:


aadi2206:
Ok Sir Sorry For That, As I thought that v1 will be main-streamed, so we should resist people to use v0 due to glitches & bugs in long run.

PR #746, opened 2025-04-30
Update Register.php

dancebunny98:
allowed only English symbols


CorwinDev:
For what is this needed?

dancebunny98:
image

CorwinDev:
That’s a problem with Pterodactyl not Paymenter

PR #748, opened 2025-05-01
Adding "backups" support for Virtfusion configurable options

nyruservices:
"backups" configurable option for Virtfusion


CorwinDev:
Lets also add a configurable option to the product settings so admins can either choose a default (or leave empty) to directly assign the backup plan

PR #749, opened 2025-05-02
Update index.blade.php

dancebunny98:
fix mobile ui

https://youtu.be/Zeee8WrMxxs


CorwinDev:
And what does this fix exactly?

dancebunny98:
> And what does this fix exactly?

The menu will be automatically closed if there was a transition to another screen resolution, and navigation is automatically hidden or shown so where necessary. (The balance is not displayed, this is already if I need to do)

image
image


image

CorwinDev:
There is no need for removal of the subtotal

PR #750, opened 2025-05-03
Update Pterodactyl.php

dancebunny98:
I delved into the code process, and realized that this was done absolutely contradictoryly all the rules and norms of the code. This is the simplest and work option that will completely remove the error when creating a user, regardless of the user registration language. In FIRST and LAST NAME! The mail all will be true in the name of the login and it will be possible not to be exhausted above the code!

### DIFF
--- a/extensions/Servers/Pterodactyl/Pterodactyl.php
+++ b/extensions/Servers/Pterodactyl/Pterodactyl.php
@@ -268,7 +268,7 @@ class Pterodactyl extends Server
         if (!$user) {
             $user = $this->request('/api/application/users', 'post', [
                 'email' => $orderUser->email,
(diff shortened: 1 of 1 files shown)


VXsz:
> In the current state it only uses the first part which is not unique (#752)
Randomly generated numbers at the end would suffice. I think using the first part of the email is better as that feels more suited towards a username, rather than a potentially long username that is currently used.

dancebunny98:
> > In the current state it only uses the first part which is not unique ([#751 (comment)](https://github.com/Paymenter/Paymenter/issues/751#issuecomment-2849289075))
> 
> Randomly generated numbers at the end would suffice. I think using the first part of the email is better as that feels more suited towards a username, rather than a potentially long username that is currently used.

Made editing namedly with the moment you are talking about

CorwinDev:
Solved in 6f2ca5e4d062ba1993cb69008bda648e66f1d835

PR #752, opened 2025-05-04
Add RTL Tag if Lang is Set to Arabic

VXsz:
This makes the direction of the pages turn to the right direction, if the language is set to Arabic. I've thought about adding more checks for other RTL languages, but since there's no one to actually test them it wouldn't be good.

I've tested the changes and dug around for bugs. I fixed the sidebar in the dashboard, which gets stuck to the left instead of actually going right due to the `fixed` tag, which I've removed, and there seems to be no reason/difference for it after checking and reading for a bit.


dancebunny98:
Can you just add your localization? There is no place for this.

CorwinDev:
Can we make this a little bit more professional by adding all RTL languages in the config/app.php and then checking this in the layout blade?

VXsz:
Sure, I'll check it out later (probably tomorrow). Not too familiar with php.

VXsz:
> adding all RTL languages in the config/app.php 

Are you sure you want me to add untranslated langs in the config/app.php?

PR #758, opened 2025-05-09
Update File "Virtualizor.php" with the upgradeServer

ExpRobi:
I created the function for upgradeServer in order to have the upgrade function for virtualizor. it is already possible to put it in a commit.

Attention: we do not know why it does not work with the upgrade by the client but only by the admin through the trigger in his panel

### DIFF
--- a/extensions/Servers/Virtualizor/Virtualizor.php
+++ b/extensions/Servers/Virtualizor/Virtualizor.php
@@ -348,7 +348,7 @@ class Virtualizor extends Server
         return 'https://' . $this->config('ip') . ':' . $this->config('client_port') . '/' . $response['token_key'] . '/?as=' . $response['sid'] . '&svs=' . $properties['server_id'];
     }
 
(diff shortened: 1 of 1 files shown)


CorwinDev:
You sure it doesn’t work when running from client?

ExpRobi:
Yes, it does not execute automatically. But if the admin goes to the upgradeServer trigger it works

CorwinDev:
I think it only does not run when the upgrade is =<0, can you check that?

ExpRobi:
I have tried your modification. and it still doesn't work. the problem arises that the client upgrade works on the graphics side. That is, if a VPS S is to be upgraded the customer pays and graphically in the panel comes out VPS M, but does not really upgrade on virtualizor. As soon as an admin performs the trigger instead it works immediately

CorwinDev:
Try the change given in 27d759b5e2a1fd394490851af7feeb7b328b322e, that *should* fix it

PR #760, opened 2025-05-09
Add Italian Translate

ExpRobi:
I added the Italian language manually. the translation is also on the crowdin site but this one is more precise.


CorwinDev:
We do not take translations via GitHub, please submit via Crowdin

PR #776, opened 2025-05-17
Use Mpdf instead of DomPDF for Better Language Support

VXsz:
Mpdf seems to be a lot better in terms of different languages support, with good defaults for fonts and such. 
DomPDF puts `?` for every foreign char, which makes the billing panel critically unusable in regions that use languages whose chars are special/unique.

I've also went ahead and added support for adaptive RTL document direction depending on the currently set language by the user, this is also great if the user wants the invoice in a different language. Also fixed column headers direction both in the rendered invoice pdf & the dashboard's invoices page, it will also change depending direction status.

A couple of notes
this may need further testing as a it's a core change, and also updating needs testing, not sure if it will break if composer doesn't install the laravel mpdf lib in the new composer.json when a user updates.

Also, there are no translations for the "Paid" text (in the invoice pdf, in front of status), or the "Download PDF" button which would be great to have translated strings for.

One last thing, I've done a cleaner job here than previously mentioned in the discord support thread, I wasn't using a laravel mpdf lib and I think it would've much more manual work.

### DIFF
--- a/app/Classes/PDF.php
+++ b/app/Classes/PDF.php
@@ -3,13 +3,13 @@
 namespace App\Classes;
 
 use App\Models\Invoice;
-use Barryvdh\DomPDF\Facade\Pdf as DomPDF;
+use Mccarlosen\LaravelMpdf\Facades\LaravelMpdf as Mpdf;
 
 class PDF
 {
     public static function generateInvoice(Invoice $invoice)
     {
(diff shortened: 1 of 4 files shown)


VXsz:
I just realized, that's definitely not how to add a new lib to composer.

CorwinDev:
Just tested MPDF but it seems that MPDF doesn't support all styles supported in DomPDF

DOMPDF:
image

MPDF:
image


I'm gonna postpone this issue for now

Issue #708, opened 2025-04-16
[Feature Request] Set max uses for coupon per customer


Davido8316:
It would be good if there was an option to set a maximum number of redemptions per customer for the coupons.

PR #789, opened 2025-05-20
feat: Add per-user coupon usage limits

Jdaepp:
Added the Feature of coupon usage limits requested by #708

Maybe i should also add that Admins can remove used coupons of the user or decrease / increase coupon usages?


Jdaepp:
Couldn't test out the lates commit as i don't have access to a Linux device. Will test it in 5-6h

Jdaepp:
Hmm, okay. I'il actually try to redo the entire thing then because i'm also not that happy with the current result.

PR #790, opened 2025-05-22
feat: add max per-user coupon limits

Jdaepp:
requested by #708 | Second try after #789

Here, again without the possibility of removing coupon usage from the User in the admin panel. Currently don't have an idea how you could implement that with the current solution.

Tested out and seems to work


Jdaepp:
Just looked at the current Service page. You can already manually remove coupons from the services so maybe it doesn't need more implementation, right?

CorwinDev:
What about if a user is logged out and signs in after adding the coupon?

Jdaepp:
Quickly tested it out and ye thats an issue. I'il try to fix it later today, currently don't have any idea. Maybe you have one?

CorwinDev:
Lets move the checks to a seperate function and call that function again directly after someone presses on checkout, if it fails, remove the coupon

PR #792, opened 2025-05-22
Docker-Compose Changes

michaelortnerit:
Hello! A little update to the docker-compose files. 

- version: '3.8' is obsolete we can see this here: https://docs.docker.com/reference/compose-file/version-and-name/

I have updated the MariaDB image to use the lts tag. This tag refers to the long-term support version, which is stable and includes the latest security patches. It works reliably in my environment, and I recommend we use it to ensure better maintainability and up-to-date support.

If you have any questions please write me! Thank you.

### DIFF
--- a/docker-compose.beta.yml
+++ b/docker-compose.beta.yml
@@ -1,4 +1,3 @@
-version: '3.8'
 x-common:
   database:
     &db-environment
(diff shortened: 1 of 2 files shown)


michaelortnerit:
I added some Readme changes. They are optional.

CorwinDev (approved):
LGTM

PR #796, opened 2025-05-22
Max coupon uses per user

Jdaepp:
requested by #708 | third try after #789 and #790 (sorry, i had some issues with github, won't close the requests anymore)

The last issue was that non logged-in Users were able to validate a coupon, log in and by that bypass the max uses check.

I now changed that when validating a coupon the ValidateCoupon function gets run and then a second time during the checkout, procedure. The second check happens after having checked if the user is logged in so it should be fixed now, right?


app/Livewire/Cart.php
```diff
@@ -117,6 +117,27 @@ public function checkout()
             return $this->notify('You must accept the terms of service', 'error');
         }
 
+        // Re-validate coupon if one exists
+        if (Session::has('coupon')) {
+            try {
+                $coupon = Session::get('coupon');
```
CorwinDev, line 123: Please move all logic in here to a seperate function in cart.php, and call removeCoupon instead of removing the coupon manually

CorwinDev (changes requested):

app/Livewire/Cart.php
```diff
@@ -117,6 +117,13 @@ public function checkout()
             return $this->notify('You must accept the terms of service', 'error');
         }
 
+        // Re-validate coupon if one exists
+        if (Session::has('coupon') && !\App\Classes\Cart::validateAndRefreshCoupon()) {
```
CorwinDev, line 121: Please just call: \App\Classes\Cart::validateAndRefreshCoupon() and if it throws a displayexception just throw that

Jdaepp:
Quickly tested it out. Works perfectly fine and all. Thx for the help :)

Issue #766, opened 2025-05-13
[Feature request] Mail Test and Massmail


michaelortnerit:
We can configure the Mail option, but we have no option where we can test if the mail is working. Maybe it would be great to make a test button, where you can send a mail to the mail address of the current logged in Administrator.

The next thing i would see on paymenter would be a massmails option where you can send massmails to all customers with maybe some configuration:

As an example

If the user is in country Austria and the customer has product x and product x is on server then send mail with following text. 

Maybe with a date option, when the job will run.

Thank you!

Jamesking56:
Might be better to split this into two issues, one to add the test button and one to add mass mailing, as I feel mass mailing is a much bigger sized issue than the test button

Somnathcoder05:
Much needed feature.
I have added SMTP on email section. But when new user register the verification mail never arrive. But unfortunately I can't test what's the problem and there's no log showing

PR #798, opened 2025-05-22
feat(admin): add test email functionality to mail settings

Jdaepp:
partly adds #766 

There are a couple of Thoughts i still have that may need changes:
1. I created a new Testmail, which you might not want. I could also just make a new Email Template and in that case use the default mail. *or use a already existing template but that would be kinda weird
2. when pressing the send test mail button it silently saves the form again. Idk if i should revert the changes so it always runs normal (with the notification of the saving)

tested out and works perfectly fine


CorwinDev:
I don't really like this approach, I'd rather use a normal template.

And I was already working on this feature :D

PR #815, opened 2025-06-06
Update active-users.blade.php

QKing-Official:
Added click to copy
Speaks for itself
image


QKing-Official:
For name, ip and mailadress.

CorwinDev:
I don't think we should remove the link to the user itself, please add that back and only add copying for the ip

QKing-Official:
> I don't think we should remove the link to the user itself, please add that back and only add copying for the ip

Sure, i will do that!

QKing-Official:
Fixed it, just copied the copy element over to old widget

PR #816, opened 2025-06-06
Update active-users.blade.php

QKing-Official:
Fixed my previous pull request to meet Corwin's requirements


CorwinDev:
Please don't make 2 pull requests but rather update your first one, also, please do not edit the layout/design of those components (so only adding copy element thingy)

QKing-Official:
> Please don't make 2 pull requests but rather update your first one, also, please do not edit the layout/design of those components (so only adding copy element thingy)

Sorry, closed previous one already, i shall work on it with copy element only without any layout changes

PR #823, opened 2025-06-17
Added UddoktaPay Gateway

rtraselbd:
Added UddoktaPay gateway for Bangladeshi users.


CorwinDev:
We only include a certain set of gateways in the core code, you're more then welcome to upload this to [BuiltByBit](https://builtbybit.com/resources/categories/paymenter.76/)

PR #824, opened 2025-06-17
Again an update on the Active user widget

QKing-Official:
Added an live user counter, just not even a one-line change.
But maybe nice i think.
image


QKing-Official:
Forgot something....

CorwinDev:
Limit 5000 doesn't seems like a good idea, please change it to just a seperate query

QKing-Official:
I think this is what you meant

CorwinDev:
Solid solution! Thanks for your pr

QKing-Official:
np

PR #828, opened 2025-06-20
fix: refactor CustomPropertyResource to dynamically load models with HasProperties trait

miwone:
This PR refactors the CustomPropertyResource to automatically discover and load models that use the HasProperties trait, instead of hardcoding them


CorwinDev:
I don't think admins should be able to make CustomProperties for all models, as those forms are not shown to the client. 

HasProperties on the services is mostly used for either manually adding a note or extension actions

PR #833, opened 2025-06-24
white unreadable text color in service and currency on navbar

Adekabang:
- Fix the white text on the white mode (background) in the service view if the service has multiple views
- Fix the currency display on the navbar by reloading the currency if changed


CorwinDev:
If we're also using a redirect, we can also remove the `CurrencyChanged` trait.

Can you revert files you did not edit? Seems like your editor formated some files

Adekabang:
Done, reverted the editor format and removed the `CurrencyChanged` trait.

CorwinDev:
LGTM

PR #839, opened 2025-06-30
fix: add closeCancel method and update button action in Show component

miwone:
fixed error `Unable to call component method. Public method [openModal] not found on component` when closing cancel modal


CorwinDev:
What a timing, I just fixed it 🙈

PR #846, opened 2025-07-02
Adding support for Hypervisor-v2 Control Panel

Bharat-B:
Hypervisor provides enterprise-grade KVM virtualization management with an intuitive web interface. Deploy, manage, and monitor virtual machines with ease.

https://hypervisor.io/

For any further information regarding the same our team is reachable at support@readydedis.com


Jdaepp:
You should probably rather make an extension for it and upload it to builtbybit

PR #847, opened 2025-07-04
Prevent password reset emails for admins

cst1x:


CorwinDev:
I do like this feature but I'd like to have a alternative in the form of a CLI command.

cst1x:
I've implemented an alternative via CLI, please check

CorwinDev:
LGTM, Why did you remove the redirect though?

cst1x:
because the notification disappears very quickly and the user won't be able to see it. also, I think that this redirect is unnecessary

CorwinDev:
Please use app:user prefix and use promptForMissingArgumentsUsing to prompt for missing email.

cst1x:
better?

CorwinDev:
Please remove everything from the return statements, there is nothing done with em, and remove the ones not needed

cst1x:
done

CorwinDev:
The ratelimiter is incorrect, this should a unique key based on the request ip

cst1x:
^

PR #858, opened 2025-07-11
Show sessions page for admin

cst1x:


CorwinDev:
Sessions are only there for 2h then they are not visible anymore, not needed in my opinion

PR #859, opened 2025-07-11
(feature) Send password reset email in admin panel

cst1x:


CorwinDev:
Users can already reset their password themselves, no need to add another button for it in admin (and if a admin really wants to send a reset password they can use the client side)

cst1x:
why not? it's easier this way as admins won't have to logout/access an incognito page to reset a user's password

CorwinDev:
When will a admin need to send a email to the user for password reset? If a user forgot their password they can just reset it themselves

PR #867, opened 2025-07-12
Update dashboard.php Polish language

bajtz:


CorwinDev:

PR #877, opened 2025-07-19
Refactor SocialLoginController to support dynamic OAuth providers and user auto-creation

miwone:
This PR modernizes the `SocialLoginController` by:

* Refactoring the OAuth logic to be dynamic and provider-agnostic via a config array
* Removing repetitive logic for each provider (DRY)
* Implementing automatic user creation if no matching email is found
* Adding proper user property association for each provider (e.g., Discord ID)
* Wrapping the login flow in a DB transaction with proper error logging and fallback
* Supporting better email/name extraction from each OAuth response
* Improving maintainability and reducing duplication across provider logic

This change preps the controller for easier future provider integration and adds more resilience in handling unexpected failures or missing data.

### DIFF
--- a/app/Http/Controllers/SocialLoginController.php
+++ b/app/Http/Controllers/SocialLoginController.php
@@ -3,7 +3,11 @@
 namespace App\Http\Controllers;
 
 use App\Models\User;
+use Illuminate\Http\RedirectResponse;
+use Illuminate\Support\Str;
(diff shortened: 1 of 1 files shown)


CorwinDev:
I don't really like this as you should not be able to create a account without filling in the properties required by admin

And I'll probably rewrite this social login to make it linking to a account instead of fixed to users email

PR #891, opened 2025-08-02
Update invoices.php

faceliftben-de:
Edit Bill_to and issued_to


CorwinDev:

Issue #904, opened 2025-08-04
[Bug]: Sliders not rendering right steps


Jdaepp:
### Checklist

- [x] I have updated to the latest available Paymenter version.
- [x] I have cleared the cache of my browser.
- [x ] I have tried a different browser to see if it is related to my browser.

### Describe the problem you are having

Sorry for the next bug report.
I have already tried changing themes and it doesn't help.

<img width="1446" height="485" alt="Image" src="https://github.com/user-attachments/assets/559c32c5-5274-400c-b692-b769d3688ab6" />

Sliders sometimes don't render right. I've already had them rendering partly right:

<img width="1355" height="525" alt="Image" src="https://github.com/user-attachments/assets/88c47e45-502f-4420-8b73-fa775bb4d272" />

And sometimes even completely correct:

<img width="1297" height="299" alt="Image" src="https://github.com/user-attachments/assets/a9c7094e-e637-4313-96f8-2e6f2665ad90" />

But now all of them are broken just like in screenshot 1
I really can't explain why and how that happens.

### Steps to reproduce

Honestly i have no clue how you can make the bug happen. Maybe a me issue again?

### Version

1.2.8

### In which browser(s) are you experiencing the issue with?

Happens in both Chrome and Edge

### Relevant Paymenter log output

```shell
-
```

### Operating system

Linux

PR #905, opened 2025-08-04
Fix: Sliders not rendering correctly in #904

Jdaepp:
I fixed this using ChatGPT so absolutely no worries if you don't want those changes. That's the result:
image


Jdaepp:
Reverted some unrequired changes

PR #909, opened 2025-08-08
Update DirectAdmin.php

dev-ricardoneud:
I fixed a security issue in a direct admin extension for the SSO login. I modified the API call for just that API call so that it actually generates the key for the user itself, because now he still logged in with the admin account


CorwinDev:
You have removed all comments, are not using the parsed argument for api requests

ghost:
I will fix the issues

PR #912, opened 2025-08-10
Implement Dynamic & Interactive Registration Form

WallyTheBot:
This pull request overhauls the user registration process by introducing dynamic and interactive form logic. The primary goal is to improve the user experience through real-time validation and to make the form's requirements more flexible based on application settings.

This resolves a usability issue where the submit button could be enabled prematurely and makes the registration process more robust.

### Key Changes & Enhancements:

* **Conditional Terms of Service (TOS) Checkbox**
    * The TOS checkbox is now only rendered if a TOS URL is configured in `config('settings.tos')`.
    * The back-end validation rule (`'terms' => 'accepted'`) is now conditionally applied, making the TOS agreement truly optional if not configured.
    * The checkbox label and link text are now fully translatable using Laravel's localization files (`lang/{locale}/auth.php`).

* **Intelligent Submit Button Control**
    * The "Sign Up" button's state is now controlled by the `isFormInvalid()` computed property in the `Register.php` Livewire component.
    * The button is only enabled when **all** required fields are filled out and the TOS (if visible) are accepted.

* **Real-Time User Feedback**
    * All input fields now use `wire:model.live`.
    * This provides instant feedback to the user, as the submit button's state updates with every keystroke, clearly indicating when the form is ready for submission.

These changes result in a more intuitive and robust registration process and improve the maintainability of the component by centralizing the validation logic.

### DIFF
--- a/app/Livewire/Auth/Register.php
+++ b/app/Livewire/Auth/Register.php
@@ -8,6 +8,7 @@ use App\Models\User;
 use App\Traits\Captchable;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Hash;
+use Livewire\Attributes\Computed;
 
 #[DisabledIf('registration_disabled')]
 class Register extends ComponentWithProperties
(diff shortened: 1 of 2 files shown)


CorwinDev:
I do not think we should make inputs live, but I do think adding the TOS acceptance to the register page is a great idea!

Please submit a different pull request with only the ToS checkbox (use like here: https://github.com/Paymenter/Paymenter/blob/master/themes/default/views/cart.blade.php#L107-L112)

PR #916, opened 2025-08-12
DM reminders with discord bot

ExpYTubuntu:
Added a bot token option in the settings and added a discord id option in the personal details tab of a user. The discord bot, of course, has to be in the same discord server as the user to work, and it HAS to be a discord id. But its a base that works flawlessly so far. I could attempt to make it work with the discord sign in option instead, but i think this is better as it allows users that aren't signed in with discord to also get dm reminders. Maybe something that gets the user id  from their discord username could also work better. Let me know if you have any concerns please

Screenshot 2025-08-12 155901
Screenshot 2025-08-12 155927
Screenshot 2025-08-12 160147
Screenshot 2025-08-12 160153

### DIFF
--- a/app/Classes/Settings.php
+++ b/app/Classes/Settings.php
@@ -492,6 +492,22 @@ class Settings
                     'default' => 'wavatar',
                 ],
                 [
+                    'name' => 'discord_dm_enabled',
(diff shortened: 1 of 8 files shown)


CorwinDev:
Not a thing planned for core code

PR #918, opened 2025-08-15
TOS Checkbox for the Register page

WallyTheBot:
https://github.com/Paymenter/Paymenter/pull/912#issuecomment-3188453652

here only checkbox addets


CorwinDev:
ToS is not required ALWAYS, only if settings.tos is set

CorwinDev:
Yup looks good! Now only thing left is to remove the German comments (or translate them into english) and revert the lines removed/added

CorwinDev:
Yup, I'll merge this PR in a little bit

PR #934, opened 2025-08-25
fix: tax calculation

AV3RG:
If you have tax set as exclusive when you add a product to cart when logged out and then login, it keeps the total price as the same and uses taxes as inclusive.

This PR solves that and any similar issues that may occur


CorwinDev:
Superseded by #972

PR #941, opened 2025-08-29
feat: Add HestiaCP integration

moshekv:
Implemented integration to HestiaCP.
Originally taken from [hestiacp_paymenter](https://github.com/hestiacp/paymenter), and fixed by me.

How to use?
First, you configure HestiaCP server:
image
(You can check config is good by clicking "Test connection")

Then in product page, just choose the package you want to sale:
image

Then after client purchase, Username and Password will appear on the service page, alongside with button directing to HestiaCP control panel.
image


CorwinDev:
We don't really accept server extensions that we can't maintain ourselves, you're more then welcome to upload it to BBB or GitHub

PR #947, opened 2025-09-03
Only show published announcements on page

Easylemon24:
Logic was missing to only show published/active announcements.


CorwinDev:
This is already being done here: https://github.com/Paymenter/Paymenter/blob/master/extensions/Others/Announcements/Announcements.php#L77

Easylemon24:
Hey, it definetly seems like it but after i tested it they still showed up on the page in the paymenter instance. I think it might be because the variable gets overriden here: https://github.com/Paymenter/Paymenter/blob/e7c22e5c47e333abf3c76d6f6ba3c8abe299a0e1/extensions/Others/Announcements/Livewire/Announcements/Index.php#L20

CorwinDev:
You're right, fixed and co-authored in 044b430a0ab168f3fd6e69372684d74b75c5080a

PR #958, opened 2025-09-12
Add Authentik SSO integration with Socialite

SwadowMaster:
Introduces Authentik as a new OAuth2 SSO provider, including settings, controller logic, a custom provider, and UI updates. Adds required dependencies and assets, updates SocialLoginController to support Authentik, and adjusts the login view to display the Authentik login option. Also registers the necessary event and service providers for Socialite and Authentik.

feat: Add Authentik OAuth2 provider integration

- Updated providers.php to include SettingsProvider and SocialiteProviders\Manager\ServiceProvider.
- Updated composer.lock with new dependencies and their versions.
- Modified login.blade.php to support Authentik as an OAuth provider.
- Created SocialiteController for handling OAuth redirects and callbacks.
- Implemented AuthentikProvider for managing Authentik OAuth2 authentication.
- Registered AuthentikProvider in EventServiceProvider for Socialite integration.
- Added Authentik SVG assets for branding.


CorwinDev:
I think you're better off submitting this as a addon (to BuiltByBit) and not as core code

SwadowMaster:
here is the extension maybe you can add it to extension page in paymenter! 
https://builtbybit.com/resources/authentik-sso-paymenter.77079

PR #962, opened 2025-09-15
Added support for Google Analytics to follow a payment through

lolbraa:
As we worked with Google to implement Analytics on our platform, they asked us to pass a URL-parameter when an invoice is marked as paid. This pull request implement a property, "paid", which is updated when the polled function checkPaymentStatus() sees the status updated. The redirect is done as a refresh so Google's JS catches the update.

This worked well for us with the Tag Manager-script put in header of app.blade.php of a custom theme.

### DIFF
--- a/app/Helpers/ExtensionHelper.php
+++ b/app/Helpers/ExtensionHelper.php
@@ -159,6 +159,10 @@ class ExtensionHelper
                 continue;
             }
 
+            if (!file_exists($path) || !class_exists($class)) {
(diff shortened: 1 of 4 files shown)


CorwinDev:
Thanks for your contribution to Paymenter!

I don't think this is really neccesary to add a url parameter to the invoice pages but rather you should be able to use [this](https://support.google.com/tagmanager/answer/7679410?hl=en)

PR #964, opened 2025-09-15
Fix possible issued with CVE-2025-58752  and CVE-2025-58751 in open Dev Environments

GomorrhaDev:
It is quite likely that it would never have become a real problem, as probably no one makes their dev environment publicly accessible. However, it's still important to prepare for the extremely unlikely (and foolish) scenario. (Especially if a significant proportion of users are under 18 and have 0 idea of what their doing).


CorwinDev:
I appreciate the PR but this has already been done in https://github.com/Paymenter/Paymenter/pull/953

PR #973, opened 2025-09-29
Feature: Display Service Hostname in Invoices and Service Listings

Adekabang:
This pull request introduces the concept of a "hostname" for services, making it visible across various parts of the application, including invoice views and service listings. The main change is the addition of a computed `hostname` attribute to the `Service` model, which is then displayed in relevant UI components if available.

**Backend functionality:**

* Added a computed `getHostnameAttribute()` method to the `Service` model, which retrieves the `hostname` (or similar `domain`/`server_name`) from service configs or properties, providing a unified way to access a service's hostname.

**Frontend/UI updates:**

* Updated invoice-related views (`pdf/invoice.blade.php`, `themes/default/views/invoices/index.blade.php`, `themes/default/views/invoices/show.blade.php`, `themes/default/views/invoices/widget.blade.php`) to display the service hostname under each relevant invoice item, if available. [[1]](diffhunk://#diff-17ea46387727661f2c57c3e0bcd2e65767f297fd7a681c64c75501ca3c8976cdL156-R172) [[2]](diffhunk://#diff-ff851c4d25d6546c63ad7a1c7b249bed5cc4392c899bf9e146b252dafe37c0e2R34-R47) [[3]](diffhunk://#diff-459a940a67cd956f8f9da4430e0c5c064d1466d44bbb7fe3e9a79ef8441bd22dR129-R143) [[4]](diffhunk://#diff-39f4cb1dfc868b94f6d348d08c26bc3a29a55c7d486a9d8708ed91cb1e1d7f86R33-R46)
* Enhanced service listing and widget views (`themes/default/views/services/index.blade.php`, `themes/default/views/services/show.blade.php`, `themes/default/views/services/widget.blade.php`) to show the service hostname alongside other key service details. [[1]](diffhunk://#diff-86dd4c6964c7d8685f742e301b96b32d3abf1912e51d92d3fa767efc7d3ba3faL32-R46) [[2]](diffhunk://#diff-7f6c14f47aa166d51f8c2b4fae9ac7042802c06b9709e1a6afb6dec80f6f59b1R58-R66) [[3]](diffhunk://#diff-c847152914fb6c448d21b56d84de90960ee19984875c6eb0c7f0d3aa6ee57431L29-R43)

**Localization:**

* Added a new translation string for "Hostname" in the English services language file.

**Service Card:**
image

**Invoice Card:**
image

**Invoice Details:**
image

**Invoice PDF:**
image

### DIFF
--- a/resources/views/pdf/invoice.blade.php
+++ b/resources/views/pdf/invoice.blade.php
@@ -153,7 +153,23 @@
         <tbody>
             @foreach($invoice->items as $item)
             <tr>
-                <td>{{ $item->description }}</td>
+                <td>
+                    {{ $item->description }}
+                    @if(in_array($item->reference_type, ['App\Models\Service', 'App\Models\ServiceUpgrade']))
+                        @if($item->reference_type == 'App\Models\Service' && $item->reference)
+                            @php
+                                $hostname = $item->reference->getHostnameAttribute();
+                            @endphp
+                        @elseif($item->reference_type == 'App\Models\ServiceUpgrade' && $item->reference && $item->reference->service)
+                            @php
+                                $hostname = $item->reference->service->getHostnameAttribute();
+                            @endphp
+                        @endif
+                        @if($hostname)
+                            <br><small style="color: #666;">{{ __('services.hostname') }}: {{ $hostname }}</small>
+                        @endif
+                    @endif
+                </td>
                 <td>{{ $item->quantity }}</td>
                 <td>{{ $item->formattedPrice }}</td>
                 <td>{{ $item->formattedTotal }}</td>
--- a/themes/default/views/invoices/index.blade.php
+++ b/themes/default/views/invoices/index.blade.php
@@ -31,6 +31,20 @@
         </div>
         @foreach ($invoice->items as $item)
             <p class="text-base text-sm">Item(s): {{ $item->description }} ({{ __('invoices.invoice_date')}}: {{ $invoice->created_at->format('d M Y') }})</p>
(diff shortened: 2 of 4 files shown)


CorwinDev:
I do like the idea of adding a unique identifier to the services but I don't really like that its fully guessing a field to be the hostname.

Maybe we can better add the unique id or something? Happy to hear your suggestions, but I don't think just guessing the properties is the way to go

Adekabang:
Yeah, I agree with you about the guessing part. Instead of trying to guess property names with like '%hostname%', I've an idea to implement it using a service identifier field approach:

approach:
  - new dedicated service_identifier database field with a unique constraint
  - auto-generation (e.g. SRV-123-ABC12345) or manual entry (hostname/domain/etc)

What do you think? Does this address your concern?

CorwinDev:
I like that, what about make a identifier that is editable by the user? So not strictly hostname/domain name but just the ID or a custom filled in identifier

Adekabang:
That makes sense. Also, the field should not be a unique value.

What about backward compatibility with the current extension? I've noticed that Convoy or Virtfusion requests the hostname during checkout. My current suggestion for existing data is to leave it empty, but we also need to update the extension to populate the data based on the hostname or domain input.

CorwinDev:
I'd just leave it empty indeed, and add a setting to the admin side to allow editing that variable as user or not 🤔?

Do you want to make this feature or should I pick it up?

CorwinDev:
Superseded by b1a7fcc6267fe2e87c4b7d275bb973c8dbcf40d8

PR #998, opened 2025-10-25
fix: bundle EasyMDE locally & resolves initial white flash on ticket forms

26bz:
Switches from using EasyMDE as a external script to become a dev deps for the bundle. It also fixes the white flash on initial load for the ticket pages that use EasyMDE.

A image of the old pure white flash can be seen below.

Screenshot From 2025-10-25 15-48-08


CorwinDev:
@26bz how big is easymde? Might be more preferable to make it a separate js file and using @vite() to include it into the ticket pages

26bz:
@CorwinDev your right, done.

PR #999, opened 2025-10-26
feat: align form component labels above inputs

26bz:
The label being placed over the inputs with a border & background are killing me. These simple changes give all labels a much cleaner look and feel more native to Paymenter.

Screenshot (8)
Screenshot (5)
Screenshot (7)
Screenshot (6)


CorwinDev:
image
I think we should remove the `mt-3` from inputs, looks a lot better

PR #1002, opened 2025-10-26
Feat: Add Alipay support for Stripe Setup Intents

gps949:
This enhancement allows users to set up Alipay as a payment method.


CorwinDev:
image
Seems that Stripe doesn't really like this, is this even a thing?

gps949:
I’m able to add a payment method without any issues on my side. Could it be related to the settings on Stripe’s side?

image

CorwinDev:
I guess so yes, so for that I don't think its a good idea to merge this

PR #1012, opened 2025-10-31
fix(docker): Prevent settings reset on container restart

gps949:
This PR fixes an issue where user settings were reset to default on every container restart.

The `entrypoint.sh` script was running `php artisan migrate --seed` on every startup. This has been changed to a conditional seed: the database is now only seeded if the `settings` table is empty. This ensures data is only seeded on the first run, preserving user configurations on subsequent restarts.

### DIFF
--- a/.github/docker/entrypoint.sh
+++ b/.github/docker/entrypoint.sh
@@ -58,7 +58,16 @@ chown -R nginx:nginx /app/storage
 
 ## make sure the db is set up
 echo -e "Migrating and Seeding D.B"
-php artisan migrate --seed --force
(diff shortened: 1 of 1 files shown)


CorwinDev:
But the setting seeder does not overwrite any settings? + new default settings might need to be seeded into the db

gps949:
I did experience the settings being reset to default on container restart, and this commit fixed the issue for me.
But you’re right — in theory, seeding shouldn’t cause that. I’ll close this PR for now, and we can revisit it if the issue appears again in the future.

PR #1017, opened 2025-11-01
Fix: Prevent duplicate execution of scheduled tasks

gps949:
This PR adds ->onOneServer() to the core scheduled tasks (ScheduleHeartbeatCommand, CronJob, FetchEmails) to prevent them from running on multiple servers simultaneously in a load-balanced environment.

This is crucial to avoid race conditions and duplicate actions, such as sending multiple invoices or fetching the same support emails, ensuring the application runs reliably when scaled horizontally.

### DIFF
--- a/routes/console.php
+++ b/routes/console.php
@@ -7,9 +7,9 @@ use App\Console\Commands\ScheduleHeartbeatCommand;
 use App\Console\Commands\TelemetryCommand;
 use Illuminate\Support\Facades\Schedule;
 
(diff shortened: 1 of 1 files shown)


CorwinDev:
Are you running it on multiple servers? 👀

gps949:
> Are you running it on multiple servers? 👀 

Yeah, I run it on two servers — mainly for better uptime, and to do A/B testing while rolling out updates without downtime. Might add more PoPs later to cut down global latency.

CorwinDev:
Thats cool, LGTM

PR #1021, opened 2025-11-01
Update Pterodactyl.php

WallyTheBot:
It is now possible for a user to select a docker image of the eggs when checking out and creating a package.


CorwinDev:
Please make docker_image optional and only applied via properties

CorwinDev:
Not as setting in getProductConfig, just check if $properties['docker_image'] is set and use that then

CorwinDev:
Fixed in referenced commit

PR #1022, opened 2025-11-01
Update show.blade.php

WallyTheBot:
added translation support for action buttons like pterodactyl go to server button


CorwinDev:
Not planned atm

PR #1026, opened 2025-11-02
feat: Allow Users to Cancel Pending Orders and Services

gps949:
This PR introduces a new feature allowing customers to cancel pending services and their associated invoices directly from the client area. This enhances user autonomy and ensures accurate stock management.

### Key Changes:

- Cancel from Service Page: A "Cancel" button is now available on the service details page if the service's status is `pending`.

- Cancel from Invoice Page: A "Cancel" button is added to the invoice page for `pending` invoices, visible only if all associated services are also `pending`.

- Backend Logic:

  - Implemented `cancelPendingService()` in `Show.php` to handle individual service cancellations.

  -  Implemented `cancelOrder()` and a corresponding `isCancellable()` check in `Show.php` for cancelling from the invoice.

- Stock Restoration: The logic ensures that when a service is cancelled, the stock for the corresponding product is incremented, preventing inventory discrepancies from abandoned orders.

### Motivation:

Currently, customers cannot cancel an order they've made but haven't paid for. This can lead to support requests for simple cancellations and can cause product stock to be reserved unnecessarily. This feature empowers users and automates the process, improving efficiency for both customers and administrators.

### DIFF
--- a/app/Livewire/Invoices/Show.php
+++ b/app/Livewire/Invoices/Show.php
@@ -32,6 +32,22 @@ class Show extends Component
 
     public $setAsDefault = false;
 
+    #[Computed]
+    public function isCancellable(): bool
+    {
+        if ($this->invoice->status !== Invoice::STATUS_PENDING) {
+            return false;
+        }
+
+        $services = $this->invoice->items()->where('reference_type', Service::class)->get();
+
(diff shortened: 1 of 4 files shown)


CorwinDev:
I don't really like this

PR #1027, opened 2025-11-02
Added missing discription in the Config Option

WallyTheBot:
In Paymenter, there is an option in the client theme to add a description to an option, but this is missing in the admin view where you can add it.


CorwinDev:
Please include the description under the config options to the client as well

CorwinDev:
The exact same reason that theme added it, so under the config options on the checkout page

PR #1030, opened 2025-11-03
Feat: Add Contextual 'Create Ticket' Buttons for Users and Services

gps949:
### Summary

This PR streamlines the ticket creation process for administrators by adding convenient "Create Ticket" shortcuts to the User and Service management pages. These new buttons also serve as a reliable alternative to the often buggy user search on the main ticket creation page.

### The Problem

Currently, when an administrator needs to create a ticket, they must use the global "Create Ticket" button on the main ticket list page. This workflow has two significant issues:

1.  **Inefficient Process:** It requires manually searching for the user and/or the related service.
2.  **Unreliable Search:** More importantly, the user search functionality on that page is often unreliable, frequently failing to find existing users even when the correct name is entered. This can be a frustrating blocker.

### The Solution

This feature introduces two new contextual buttons that completely bypass these problems:

1.  **On the "Manage User Tickets" Page:** A "Create Ticket" button has been added. Clicking it opens the new ticket form with the current user **automatically selected**.
2.  **On the "Edit Service" Page:** A "Create Ticket" button has been added. Clicking it opens the new ticket form with both the service's owner (user) and the service itself **automatically selected**.

These changes not only make the process faster and more intuitive but also more reliable by pre-filling the correct data, avoiding the problematic search functionality entirely.

### DIFF
--- a/app/Admin/Resources/ServiceResource/Pages/EditService.php
+++ b/app/Admin/Resources/ServiceResource/Pages/EditService.php
@@ -14,6 +14,7 @@ use Filament\Forms\Components\Checkbox;
 use Filament\Forms\Components\Select;
 use Filament\Notifications\Notification;
 use Filament\Resources\Pages\EditRecord;
+use App\Admin\Resources\TicketResource;
 
 class EditService extends EditRecord
 {
@@ -22,6 +23,13 @@ class EditService extends EditRecord
     protected function getHeaderActions(): array
     {
         return [
+            Action::make('create_ticket')
+                ->label(__('ticket.create_ticket'))
+                ->color('gray')
+                ->url(fn () => TicketResource::getUrl('create', [
+                    'ownerRecord' => $this->record->user_id,
+                    'service_id' => $this->record->id,
+                ])),
             DeleteAction::make()
                 ->form(function (DeleteAction $action) {
                     $status = !in_array($this->record->status, [Service::STATUS_PENDING, Service::STATUS_CANCELLED]) && $this->record->product->server_id !== null;
--- a/app/Admin/Resources/TicketResource/Pages/CreateTicket.php
+++ b/app/Admin/Resources/TicketResource/Pages/CreateTicket.php
@@ -11,6 +11,23 @@ class CreateTicket extends CreateRecord
 {
     protected static string $resource = TicketResource::class;
 
+    public function mount(): void
+    {
+        parent::mount();
+
+        $data = [];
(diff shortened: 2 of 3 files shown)


CorwinDev:
Doesn't really make sense to add a button for that

gps949:
> Doesn't really make sense to add a button for that

You can think it’s not elegant, that’s fine — but I really don’t see how it “doesn’t make sense.”
The whole point of this PR is that when creating a ticket from the Admin “Tickets” page using the New Ticket button, the user search often fails — it just can’t find the user by name.
And even if it works, it’s still a clumsy workflow: when an admin notices something and wants to open a ticket for a specific user or service, they have to remember the username or service ID, go back to the Tickets page, click New Ticket, then search again.

Wouldn’t it be a lot nicer if you could just create the ticket right from that user’s or service’s page directly? ；)

CorwinDev:
Its almost never that a admin makes a ticket for a client? (I think?) To add a seperate button that will almost never be used on the services page doesn't make sense then.

I'll check whats up with the ticket searching, if thats a bug I'd rather fix that then make a button on the edit services page

gps949:
> Its almost never that a admin makes a ticket for a client? (I think?) To add a seperate button that will almost never be used on the services page doesn't make sense then.
> 
> I'll check whats up with the ticket searching, if thats a bug I'd rather fix that then make a button on the edit services page

Well, first — what if an admin needs to contact a user?
Things like price adjustment notices, policy violations, or anything that needs to reach them directly. In those cases, creating a ticket for them is way better than copying their email and sending a message manually, right?

Second, if admins really never need to open tickets for users, then why even have the New Ticket button on the Tickets page at all?

And yeah, if the user search there can’t find users by name (or even by part of the name), that’s definitely a bug worth fixing. I didn’t dig into it deeply — maybe it’s related to switching to Postgres — but at the time I just needed to reach the user quickly, so I added those two buttons. Honestly, it turned out to be much more convenient than opening a blank form and filling everything manually.

CorwinDev:
Well in the rare occasion that a admin needs to open a ticket they can do so via the create button, that shouldn't be a button on the service imo


Could you check if d61efed8a8c5847e15aa99dd8f061c6609806d3a solves your issues?

gps949:
No worries, I don’t really need to spend time checking that anymore — I’ve got a better way now to fill in the user I need. Pretty sure I won’t be using that dropdown again anyway.

PR #1036, opened 2025-11-04
Add Blesta import command to Paymenter

joaquimcc:
ImportFromBlesta console command for migrating Blesta data into Paymenter,  including:

- Users 
- Admins 
- Categories 
- Products
- Currencies
- Payments / Transactions
- Tickets
- Orders
- Services
- Coupon Codes
- Includes upgrades for each service you had setup just like in blesta

Any service/product using the pterodactyl Module will have that also enabled in paymenter

Similar installation as the WHMCS import process:

### On Blesta:
1. Export your Blesta database:
   ```bash
   mysqldump -u root -p blesta > blesta.sql
   ```

### On Paymenter:
1. Create a temporary database:
   ```sql
   CREATE DATABASE blesta_temp;
   GRANT ALL PRIVILEGES ON blesta_temp.* TO 'paymenter'@'127.0.0.1' WITH GRANT OPTION;
   ```

2. Import the Blesta database:
   ```bash
   mysql -u root -p blesta_temp < blesta.sql
   ```

3. Run the import command:
   ```bash
   php artisan app:import-from-blesta blesta_temp
   ```
   - You'll be prompted for:
     - Database password
     - Blesta system_key (optional, but enables password compatibility)

4. Clean up the temporary database:
   ```sql
   DROP DATABASE blesta_temp;
   ```

Thats it

### DIFF
--- a/app/Console/Commands/ImportFromWhmcs.php
+++ b/app/Console/Commands/ImportFromWhmcs.php
@@ -875,7 +875,7 @@ class ImportFromWhmcs extends Command
 
                 $data[] = [
                     'id' => $record['id'],
-                    'number' => !empty($record['invoicenum']) ? $record['invoicenum'] : $record['id'],
+                    'number' => !empty($record['invoicenum']) ? $record['invoicenum'] : null,
                     'user_id' => $record['userid'],
                     'status' => match ($record['status']) {
                         'Paid' => 'paid',


CorwinDev:
Few things:
- I do not think that we need a different AuthServiceProvider for blesta, afaik passwords are also stored as brcypt (which Paymenter supports)
- I'd do servers importing either more general, OR like whmcs importing, don't import servers
- For example with the importing of Cancellations, you're doing a try on service_cancellations but that never exists? So please remove that try catch as it will always not exist (same thing on orders)

joaquimcc:
yeah your right, i also noticed those, will update it

PR #1042, opened 2025-11-05
feat: Knowledgebase articles

26bz:
This PR introduces the core missing feature: Knowledgebase functionality for Paymenter 🎉 

<details>
  <summary>(client)</summary>
  
  
  
  
</details>

<details>
  <summary>(admin)</summary>
  
  
  
</details>

### DIFF
--- a/app/Admin/Resources/CouponResource.php
+++ b/app/Admin/Resources/CouponResource.php
@@ -104,7 +104,12 @@ class CouponResource extends Resource
 
                 Select::make('products')
                     ->label('Products')
-                    ->relationship('products', 'name')
+                    ->relationship(
+                        name: 'products',
+                        titleAttribute: 'name',
+                        modifyQueryUsing: fn (\Illuminate\Database\Eloquent\Builder $query) => $query->with('category')
+                    )
+                    ->getOptionLabelFromRecordUsing(fn (\App\Models\Product $record) => "{$record->name} - {$record->category->name} (#{$record->id})")
                     ->multiple()
                     ->preload()
                     ->placeholder('Select the products that this coupon applies to')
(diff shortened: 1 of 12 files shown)


hy3pez:
This is awesome. 😊

26bz:
ready for feedback so we can ship

AV3RG:
I think it would be better to keep this as a separate extension rather than bundling it with Paymenter. There are already other third-party extensions that provide similar functionality, and it’s quite straightforward to install them independently. Including it in the core package doesn’t seem necessary in this case

26bz:
Just because other knowledgebase extensions are easy to install doesnt seem like a valid reason to not ship with default paymenter. The other ones are paid and all billing automation CMS should ship knowledgebase features by default

AV3RG:
I am not opposed to implementing this as an extension; however, I am against making it a native part of Paymenter. The platform already includes built-in discovery, and installation is extremely simple. Anyone who requires this functionality can easily find your extension and use it for free, provided you choose to publish it at no cost.

The argument that every billing panel must include this feature by default does not align with the current philosophy of the project. For example, even a basic Pages extension is not provided by default in the repo.

Including this feature within the core project would also place the responsibility on the maintainers to keep the extension up-to-date with future changes and patches, which is not desirable.

hy3pez:
I understand the idea of keeping Paymenter’s core lightweight, but I still think this is something users expect to be included out of the box. Competing panels, such as WHMCS and Blesta, include a knowledge base by default, and having it built in would make Paymenter more competitive and improve the overall onboarding experience. I’m also not sure about the maintenance concern. A knowledge base is generally pretty stable once implemented, so it shouldn’t introduce much ongoing work. A simple, minimal version in the core could strike a good balance without adding real overhead.

PR #1043, opened 2025-11-05
Added missing description to ConfigOption

WallyTheBot:
There is a variable for the description in the client theme, but it is empty because you cannot add a description. With this PR, this is now possible.

No Chnage need in the client theme

_I deleted the old post because I accidentally posted a lot of changes there._

### DIFF
--- a/app/Models/ConfigOption.php
+++ b/app/Models/ConfigOption.php
@@ -14,6 +14,7 @@ class ConfigOption extends Model implements Auditable
 
     protected $fillable = [
         'name',
+        'description',
         'env_variable',
         'type',
         'sort',


CorwinDev:
It makes no sense to add a field that is not used in the default theme, please add it in the default theme as well :)

CorwinDev:
Oh you're right! Will merge this in a bit :D

CorwinDev:
Apparently Corwin from a year back saw that you were gonna pull request this

CorwinDev:
I'm not gonna merge this into 1.4, I'll merge this in probably a week or so

Issue #1039, opened 2025-11-04
[Feature] Product dropdowns using Names instead of slugs


Jdaepp:
**Describe what you are trying to accomplish and why in non technical terms**
Currently, when e.g. creating a new Order or service, you have to specify what product it uses. If products from different categories have the same name (for example small, medium and large in the categories Minecraft and Factorio), the dropdown will show the names small, medium and large, which means you'll not be able to differenciate between them:

<img width="723" height="530" alt="Image" src="https://github.com/user-attachments/assets/6ce0beba-58c5-43a0-a7e9-08b5049212d3" />

As the picture shows, there are two "Small" products. Yes, it's theoretically sorted by the same sorting as in the categories tab, but it's just a pain to differenciate.

**Describe the solution you'd like**
The best way of fixing this is by using the product-slug instead of the product name, which is already part of the products properties.

**Describe alternatives you've considered**
The alternative would be to stay with the current version, but I think it would be a valuable change that could help.

CorwinDev:
Fixed in #1044

PR #1044, opened 2025-11-05
Clearer product dropdowns

WallyTheBot:
In this PR, I focus on a feature suggestion from a user.  [#1039](https://github.com/Paymenter/Paymenter/issues/1039)

I have changed the display format of the product dropdowns so that if, for example, you have two categories with the same package names, the dropdowns now show the name, slug, and ID so that you can better see where everything belongs.

### DIFF
--- a/app/Admin/Resources/OrderResource.php
+++ b/app/Admin/Resources/OrderResource.php
@@ -73,18 +73,17 @@ class OrderResource extends Resource
                         Select::make('product_id')
                             ->label('Product')
                             ->required()
-                            ->options(
-                                function () {
-                                    return Product::with('category')
-                                        ->get()
-                                        ->mapWithKeys(function (Product $product) {
-                                            $categoryName = $product->category->name;
-                                            return [$product->id => "{$product->name} - {$categoryName} ({$product->id})"];
-                                        })
-                                        ->toArray();
-                                }
+                            ->relationship(
+                                name: 'product',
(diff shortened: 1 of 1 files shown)


app/Admin/Resources/CouponResource.php
```diff
@@ -104,7 +105,8 @@ public static function form(Schema $schema): Schema
 
                 Select::make('products')
                     ->label('Products')
-                    ->relationship('products', 'name')
+                    ->relationship('products')
+                    ->getOptionLabelFromRecordUsing(fn (Product $record) => "{$record->name} ({$record->slug}) - ID: {$record->id}")
```
CorwinDev: I'd rather use the category name instead of the product slug

CorwinDev (changes requested):

app/Admin/Resources/CouponResource.php
WallyTheBot: what format we use?
Categorie - Name
Name - Category
or something else

CorwinDev: Name - Category (id)

CorwinDev:
You removed the filters again..

CorwinDev:
You still removed the class import

CorwinDev:
`OrderResource.php` is left

CorwinDev:
LGTM

PR #1059, opened 2025-11-09
Update index.blade.php

WallyTheBot:
The design has been better adapted to the other sections, such as Security, Payment Methods, and Notifications, giving it a more uniform look.


CorwinDev:
I don't really like the inputs against that background, I'll see if we can maybe change the input colors a bit so we make this more uniform

PR #1060, opened 2025-11-09
Implement custom avatar upload system

WallyTheBot:
This pull request introduces a complete system for users to upload their own custom avatars, moving away from a Gravatar-only implementation.

The feature is configurable via the Admin Panel and provides a seamless, reactive upload experience for the client.


CorwinDev:
Don't think this is a neccesary feature, you're more then welcome to upload this as extension though :)

PR #1068, opened 2025-11-11
Added CookieBanner as core extension

QKing-Official:
Added CookieBanner as core extension.
I can always try....


ExpYTubuntu:
I believe in you

CorwinDev:
Closing because Paymenter does not use optional cookies (a cookie banner is not required in that case)

PR #1072, opened 2025-11-12
Feat: add more hooks

ghost:
Adds more hooks for extensions to hook into. Allows for more content on the homepage.


CorwinDev:
Just hook into the home event again, then you can show your stuff multiple times, this doesn't need home1, home2 etc

ghost:
@CorwinDev 

You currently cant display anything between hero and services

PR #1079, opened 2025-11-15
Fix: Support HTML email replies in ticket

gps949:
## Problem
  `FetchEmails` command fails with database error when processing HTML-only emails:
  SQLSTATE[23502]: Not null violation: null value in column "body"

  ## Solution
  - Make `ticket_mail_logs.body` nullable
  - Add `getEmailBody()` helper to extract content from both text and HTML emails
  - Add validation to skip empty emails
  - Strip HTML tags and use `EmailReplyParser` to remove quoted content

  ## Changes
  - `app/Console/Commands/FetchEmails.php`: Enhanced email body extraction
  - Database migrations: Make body column nullable

### DIFF
--- a/app/Console/Commands/FetchEmails.php
+++ b/app/Console/Commands/FetchEmails.php
@@ -54,7 +54,16 @@ class FetchEmails extends Command
                     continue;
                 }
 
-                $body = \EmailReplyParser\EmailReplyParser::parseReply($email->text());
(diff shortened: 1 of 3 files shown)


CorwinDev:
In my opinion we should not make body null in database but rather fill '(No message body found)' as body while triggering a failed email, please change this and then its good to go :)

PR #1084, opened 2025-11-20
Add support for environment variable overrides

beingsuz:
* Added logic to scan `$settings` for keys starting with `ENV_`, strip the prefix, and set the corresponding environment variable in the `$environment` array, allowing overrides for server environment variables.


beingsuz:
@radisic023 😉

CorwinDev:
Environment variables are already overwritten https://github.com/Paymenter/Paymenter/blob/master/extensions/Servers/Pterodactyl/Pterodactyl.php#L267

beingsuz:
🗿

PR #1101, opened 2025-12-12
Change Cron description to 12hr

CpuSocket:
Description states its in 24hr when its actually in 12hr.


CorwinDev:
image
Is pretty much 24h for me, but probably depends on the location you're at

CpuSocket:
image
Ah fair enough, in settings tab it isnt for me.

PR #1110, opened 2025-12-24
feat(lang): add Simplified Chinese (zh-CN) language support

DragonYH:
add Simplified Chinese (zh-CN) language support


CorwinDev:

PR #1117, opened 2025-12-29
Fixes Service API request

RAB-Media:
Fixes API to allow pulling service properties from GET function.


RAB-Media:
This makes it so that the API docs work properly

PR #1118, opened 2025-12-29
Header fix to allow logos with text for company name

RAB-Media:
Allows you to turn off the company name in the header if you are using a logo that has the name or text in it.
**Before**
image
**After**
image


CorwinDev:
You already can do this, just set logo to logo-only.

RAB-Media:
I looked and didn't see a way to do it.

RAB-Media:
I found it i see this made it a global setting not just a theme setting. i did not see it because i use a theme other than the default one.

PR #1122, opened 2026-01-02
Add Docker Node Asset‑Builder Service

BluetonicBeats:
# Add Docker Node Asset‑Builder Service

This Pull Request introduces a new Node.js asset‑builder service to the Docker Compose setup.  
It provides an isolated environment for building frontend assets (themes and extensions) without requiring Node on the host system.

## Summary

- Adds a new service: `asset-builder`
- Included in:
  - `docker-compose.beta.yml`
  - `docker-compose.example.yml`
- 42 additions, 0 deletions across both files
- No existing services are modified

## New Service: `asset-builder`

```yaml
asset-builder:
  image: node:22-alpine
  profiles: ["build"]
  working_dir: /app
  volumes:
    - "./themes:/app/themes"
    - "./extensions:/app/extensions"
    - "./:/app/var"
    - "app:/app"
  command: >
    sh -c "echo 'Asset Builder ready for npm commands' &&
           echo 'Working directory: /app' &&
           echo 'Usage: docker compose run --rm asset-builder npm install && npm run build' &&
           tail -f /dev/null"
  networks:
    - paymenter_nw
```

## Purpose

The asset‑builder service is designed to:

- Separate Node-based build processes from the PHP application stack
- Provide a reproducible environment for:
  - `npm install`
  - `npm run build`
- Support building themes and extensions in a consistent, containerized workflow

## Usage

```bash
docker compose run --rm asset-builder npm install
docker compose run --rm asset-builder npm run build
```

## Files Changed

| File | Changes |
|------|---------|
| `docker-compose.beta.yml` | +21 lines |
| `docker-compose.example.yml` | +21 lines |

PR #1139, opened 2026-01-23
feat(docker): add node/npm support for theme building

beingsuz:
Added node/npm to the bin directory to allow for themes building.
this is a simple alternative to #1122 

`docker exec paymenter npm run build <themename>`
or any other way to execute command in the docker container

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,6 +38,10 @@ RUN npm run build
 # Switch back to the final stage
 FROM final AS production
 COPY --from=build /app/public /app/public
+COPY --from=build /app/node_modules /app/node_modules
+COPY --from=build /usr/local/bin/node /usr/local/bin/node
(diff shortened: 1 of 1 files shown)


BluetonicBeats:
I’m not in favor of adding build tools directly into the main application container. Mixing runtime and build‑time dependencies increases the attack surface, slows down deployments, and makes the image harder to maintain. Build steps should stay in a separate stage or pipeline so the production container remains lean, predictable, and secure.

beingsuz:
Fair enough I just feel like it should be simple to use as if they installed it like a normal setup.

CorwinDev:
> Fair enough I just feel like it should be simple to use as if they installed it like a normal setup.

On non-docker users also need to install npm manually and it isn't included as runtime-dependency

PR #1161, opened 2026-02-16
feat: Enhance Stripe customer creation

mfyll:
## Enhance Stripe customer creation

- One-time payments now always create Stripe customers (previously only billing agreements did)
- Added "Create detailed Stripe customers" toggle to control detail level 
- When enabled: customers include address, phone, and company details 
- When disabled: customers include only name, email, and user ID 

**Why:**
- Improves payment tracking in Stripe Dashboard (no more "guest" payments)
- Allows the usage of tools like Stripe Radar to reduce fraud
- Toggle allows privacy-conscious merchants to limit stored data in Stripe

### DIFF
--- a/extensions/Gateways/Stripe/Stripe.php
+++ b/extensions/Gateways/Stripe/Stripe.php
@@ -177,18 +177,16 @@ class Stripe extends Gateway
             'metadata' => ['invoice_id' => $invoice->id],
         ];
 
-        // Always create/get a Stripe customer for authenticated users, use detailed info based on setting
-        if ($invoice->user) {
-            try {
-                $includeDetails = (bool) $this->config('stripe_create_customers');
-                $customer = $this->createOrGetStripeCustomer($invoice->user, $includeDetails);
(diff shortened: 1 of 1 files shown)


joaquimcc:
It always created stripe customers? At least from my experience one time payments where 9/10 properly linked to customers unless credit card payment method was / is being used then it does always convert into a guest payment for some reason

mfyll:
> It always created stripe customers? At least from my experience one time payments where 9/10 properly linked to customers unless credit card payment method was / is being used then it does always convert into a guest payment for some reason

Right now Paymenter only creates Stripe customers if the customer decides to create a billing agreement, if the customer selects "one-time payment", a Stripe customer is never created and thus the payment gets stored without a customer (as a "guest" customer). 

bilde

The main goal with this PR however is to allow the merchant using Paymenter to enable sending more data to Stripe to make things like Stripe Radar more useful.

bilde

mfyll:
Is there any information on when there will be any movement here?

PR #1203, opened 2026-04-24
dp-core-02: Extract dynamic_slider to shared Blade partial

Jordanmuss99:


CorwinDev:
Not planned to add changes suggested

PR #1220, opened 2026-05-09
Use updated var

CpuSocket:


CpuSocket:
Changes the var to the correct and updated one for the new " Renewal date ". 
Currently it was showing the old expire date which is returning:
image

PR #1232, opened 2026-05-20
fix/use-translations

5k0z:
Added full translation of the Paymenter administrative section.
With this fix, it's possible to have the admin area synchronized with the configured language.


CorwinDev:
Can we make all admin translations under lang/en/admin? Keeps the translations better seperated.

PR #1237, opened 2026-05-24
patch EmailReplyParser

### LINKED ISSUE #1236: [Bug]: FetchEmails crashed when receiving html
### Checklist

- [x] I have updated to the latest available Paymenter version.
- [x] I have cleared the cache of my browser.
- [x] I have tried a different browser to see if it is related to my browser.

### Describe the problem you are having

Paymenter's `FetchEmails` command crashed when receiving HTML-only emails (no plain-text part) because `$email->text()` returned `null`, which was inserted directly into a `NOT NULL` database column. Additionally, even when emails were processed successfully, the ticket message body was empty or garbled because `EmailReplyParser` was being fed a null value, and quoted reply content (including encoding artifacts) was leaking into the stored message.

### Steps to reproduce

Sending an HTML email to the inbox which the ticket system is connected to did not import the email as expected.

### Version

1.5.0

### In which browser(s) are you experiencing the issue with?

_No response_

### Relevant Paymenter log output

```shell
[2026-05-24 23:25:02] production.ERROR: FetchEmails failed: SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'body' cannot be null (Connection: mysql, Host: 127.0.0.1, Port: 3306, Database: paymenter, SQL: insert into `ticket_mail_logs` (`message_id`, `subject`, `from`, `to`, `body`, `status`, `updated_at`, `created_at`) values (5c77024a-8b71-4a74-9f4e-9c27fa03ca01@email.android.com, Hjälp mig, [user email address], [support inbox email address], ?, unprocessed, 2026-05-24 23:25:02, 2026-05-24 23:25:02))
```

### Operating system

Linux

### Any other information that may be helpful

_No response_

emollusion:
Resolves #1236 

- Fall back to `$email->html()` when `$email->text()` is null, in both the success path and `failedEmailLog()`, preventing null constraint violations on `ticket_mail_logs.body`
- Strip `<blockquote>` elements from HTML before text extraction to remove quoted previous messages from the stored reply
- Strip the `On <date>, X wrote:` attribution line left behind after blockquote removal
- Use `<?xml encoding="utf-8" ?>` prefix when loading HTML into `DOMDocument` to prevent UTF-8 characters (e.g. Swedish Å Ä Ö) from being mangled
- Plain-text emails are unaffected — the HTML parsing path is only taken when `$email->text()` is null

### DIFF
--- a/app/Console/Commands/FetchEmails.php
+++ b/app/Console/Commands/FetchEmails.php
@@ -55,7 +55,17 @@ class FetchEmails extends Command
                     continue;
                 }
 
-                $body = EmailReplyParser::parseReply($email->text());
+                $rawBody = $email->text() ?? (function($html) {
+                    $dom = new \DOMDocument();
+                    @$dom->loadHTML('<?xml encoding="utf-8" ?>' . $html, LIBXML_HTML_NOIMPLIED | LIBXML_HTML_NODEFDTD);
+                    foreach (iterator_to_array($dom->getElementsByTagName('blockquote')) as $node) {
+                        $node->parentNode->removeChild($node);
+                    }
+                    $text = strip_tags($dom->saveHTML());
+                    $text = preg_replace('/On .+wrote:\s*/s', '', $text);
+                    return trim($text);
+                })($email->html()) ?? '';
+                $body = EmailReplyParser::parseReply($rawBody);
 
                 // Check headers to see if this email is a reply
                 $replyTo = $email->inReplyTo();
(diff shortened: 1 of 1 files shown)


CorwinDev:
Can't we use something much simpler like:

```
$rawText = $email->text();

if (empty(trim($rawText)) && !empty($email->html())) {
    $rawText = html_entity_decode(strip_tags($email->html()));
}
```

The `On <date>, X wrote:`, is already being removed by EmailReplyParser

emollusion:
I am in no way a pro so if you think this is a better way to solve it than I am all for it.
It becomes a question wether the content within what was quoted (the earlier reply) should be kept or not in the new message in the thread of the case?

I had some trouble with encoding so not sure we should remove that though.

CorwinDev:
Added a commit, let me know what you think @emollusion

emollusion:
@CorwinDev , looks good, I have nothing to add. 👌

PR #1242, opened 2026-05-28
Enhance memory setting logic in Pterodactyl.php

KingofMidnight:
Add memory handling for server settings with fallback. Made the ability to use {memory} and {memory}+1024 to make an Extra Server RAM Config Option Easier

Might make more edits if i think i might need it and i think i should share it if it would be even useful for others


CorwinDev:
I don't really like this approach, I'd rather setup some sort of calculations on the configOptions itself?

KingofMidnight:
well that is understandable i only just did what i can or able too do, plus my coding skill are not much so i made that and just sent it thinking to could help spark something better

CorwinDev:
Yeah I'll see what I can do non-specific pterodactyl

PR #1244, opened 2026-05-28
Fix duplicate invoice number race condition causing 500 errors

Mxmilu666:
This PR fixes a race condition in invoice number generation that can occur under high concurrency in production environments.

In production deployments, multiple invoice creation or payment events processed at the same time could generate the same invoice number. This may result in duplicate invoice IDs and eventually trigger database exceptions or HTTP 500 errors during invoice creation/payment processing.

The previous implementation relied on reading the current invoice number from config, incrementing it in PHP, and writing it back afterwards. Under concurrent workloads, this operation was not atomic and could lead to duplicate invoice numbers being assigned.

### Changes

* wrap invoice number generation in a database transaction
* use `lockForUpdate()` to prevent concurrent writes
* verify generated invoice numbers are unique before assignment
* extract invoice number formatting into a dedicated method
* improve resilience against inconsistent counter state

This change makes invoice number generation concurrency-safe and prevents duplicate invoice numbers in high traffic production environments.

### DIFF
--- a/app/Listeners/InvoiceNumberListener.php
+++ b/app/Listeners/InvoiceNumberListener.php
@@ -4,7 +4,6 @@ namespace App\Listeners;
 
 use App\Events\Invoice\Creating;
 use App\Events\Invoice\Updating;
-use App\Models\Invoice;
 use App\Models\Setting;
 use Illuminate\Support\Facades\DB;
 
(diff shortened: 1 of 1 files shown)


CorwinDev:
I don't like that we need to check if the invoice is already used, thats up to the administrator to make sure to not set it lower then the invoices are.

So please remove that section and just lock the setting field while defining the invoice number

Mxmilu666:
OK, Changed.
Thank you for your suggestion.

PR #1246, opened 2026-05-29
feat: Include renewal invoice details in service suspension notifications

Abh1jot:
## Summary

When a service is suspended due to non-payment, the suspension notification now includes the pending renewal invoice details — amount due, itemized breakdown, and a direct "Pay Invoice & Reactivate" button.

### Problem

Currently, the suspension email only says:
> "Your service has been suspended due to a payment failure. Please pay the invoice to reactivate the service."

It doesn't tell the user **which invoice**, **how much they owe**, or provide a **direct payment link**. Users have to log in, navigate to invoices, and find the right one manually.

### Solution

The `SuspendJob` now looks up the pending renewal invoice (which the CronJob has already created before suspension) and passes it to the notification template.

**Before:**
- Service name only
- No invoice info
- No payment link

**After:**
- Service name
- Invoice number and amount due
- Itemized table (same format as invoice created email)
- Direct "Pay Invoice & Reactivate" button
- Falls back gracefully when no invoice exists

### Changes

| File | Change |
|------|--------|
| `app/Jobs/Server/SuspendJob.php` | Look up pending invoice, pass to notification |
| `database/seeders/EmailTemplateSeeder.php` | Updated template with conditional invoice block |
| `database/migrations/2026_05_29_...` | Conditionally updates existing templates (only if unmodified) |

### Backward Compatibility

- ✅ **No breaking changes** — the extra data keys are simply additional Blade template variables
- ✅ **Admin-customized templates preserved** — the migration only updates templates that still match the original default body
- ✅ **Graceful fallback** — if no pending invoice exists, falls back to generic "View Invoices" link
- ✅ **No new dependencies** — uses existing models and relationships

### Testing

1. Create a service with an active recurring billing cycle
2. Let the invoice generate via cron (or create manually)
3. Trigger suspension (let service expire past the suspend threshold)
4. Check the suspension email — it now includes invoice details and payment link

### DIFF
--- a/app/Jobs/Server/SuspendJob.php
+++ b/app/Jobs/Server/SuspendJob.php
@@ -41,6 +41,19 @@ class SuspendJob implements ShouldQueue
         }
 
         if ($this->sendNotification) {
+            // Find the pending renewal invoice for this service so we can
+            // include the amount due and a direct payment link in the notification
+            $pendingInvoice = $this->service->invoices()
+                ->where('status', 'pending')
+                ->latest()
+                ->first();
+
+            if ($pendingInvoice) {
+                $data['invoice'] = $pendingInvoice;
+                $data['invoiceTotal'] = $pendingInvoice->formattedTotal;
+                $data['invoiceItems'] = $pendingInvoice->items;
+            }
+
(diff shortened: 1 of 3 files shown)


CorwinDev:
Please do not send AI generated pull requests, this pr has functionallity on the most weird places and incorrect usage.

PR #1253, opened 2026-06-03
Add `:heavy` image and fix update flow

BluetonicBeats:
## What does this PR do?

- **New `:heavy` Docker image** ([Dockerfile.heavy](Dockerfile.heavy))
  Same as the standard image, but with the full build toolchain baked in
  (node, npm, composer, native build deps, pre-installed `node_modules`).
  Lets you rebuild theme/extension assets directly inside the running
  container — no `asset-builder` sidecar needed:

  ```bash
  docker compose exec paymenter sh -c "npm install && npm run build"
  ```

- **New compose file** [docker-compose.heavy.yml](docker-compose.heavy.yml)
  using `ghcr.io/paymenter/paymenter:heavy`.

- **Updates no longer require removing the `app:` volume.**
  [docker-compose.example.yml](docker-compose.example.yml) and
  [docker-compose.beta.yml](docker-compose.beta.yml) no longer mount a
  named volume over `/app`. Only the paths that actually need to persist
  are bind-mounted (`./var`, `storage/logs`, `storage/app/public`,
  `themes`, `extensions`). The `asset-builder` sidecar (which only
  existed because of that volume) has been removed.

  Updating is now simply:

  ```bash
  docker compose pull
  docker compose up -d
  ```

- **Snapshot layout consolidated** under `/app/default/`:

  | Old | New |
  | --- | --- |
  | `/app/themes_default` | `/app/default/themes` |
  | `/app/extensions_default` | `/app/default/extensions` |
  | `/app/public/default_default` | `/app/default/build` *(heavy only)* |

  Image and entrypoint always ship together, so this is internal but
  flagged here for anyone who scripted against the old `*_default`
  paths.

## Why the changes

The `app:` named volume overlaid the entire `/app` directory of the
image, hiding new code on every pull. Users had to wipe the volume to
update, which also wiped any in-volume state. Mounting only the paths
that need persistence fixes this for both the standard and heavy
images.

## Files changed

- `Dockerfile.heavy` (new)
- `docker-compose.heavy.yml` (new)
- `Dockerfile` — snapshot dirs consolidated under `/app/default/`
- `docker-compose.example.yml` — drop `app:` volume + `asset-builder`
- `docker-compose.beta.yml` — drop `app:` volume + `asset-builder`
- `.github/docker/entrypoint.sh` — read snapshots from
  `/app/default/{themes,extensions,build}`; seed/renew
  `/app/public/default` from `/app/default/build` when present
- `.github/workflows/docker.yaml` — also build & push `:heavy`,
  `:heavy-master`, `:heavy-<tag>`

## Not yet verified

- The new GitHub Actions step that builds and pushes the `:heavy` image
  hasn't been run yet. Needs a CI run on this branch to confirm the
  multi-arch build succeeds and the tags land in GHCR.

## Persisting the default theme's build (optional)

The bundled `default` theme compiles to `/app/public/default/` (per
`themes/default/vite.config.js` → `buildDirectory: 'default/'`). By
default this directory is **not** bind-mounted, so:

- Default-theme assets always come from the image — updates "just work".
- If you rebuild it inside the container with `npm run build`, the
  compiled output is lost on `--force-recreate` or image pull.

If you want it to persist, uncomment the optional mount in
[docker-compose.heavy.yml](docker-compose.heavy.yml):

```yaml
- "./build/default:/app/public/default"
```

Convention: `./build/<themename>` on the host maps to
`/app/public/<themename>` in the container — one subdirectory per
theme, all grouped under `./build/`.

The entrypoint seeds an empty `./build/default/` from the image's
baked-in snapshot at `/app/default/build`, so the theme keeps working
on first start. On image updates the bundle is refreshed automatically
unless you set `PAYMENTER_SKIP_DEFAULT=true`.

For **custom themes** that set their own `buildDirectory` (e.g.
`'mytheme/'`), add a parallel mount such as
`- "./build/mytheme:/app/public/mytheme"`. There is no automatic
seeding for those — run `npm run build` once after enabling the mount.

## Migration notes for existing users

1. Update your `docker-compose.yml` to the new version from this PR.
2. Stop the stack and remove the old `app` named volume:

   ```bash
   docker compose down
   docker volume rm <project>_app
   ```
3. Pull and start as usual:

   ```bash
   docker compose pull
   docker compose up -d
   ```

### DIFF
--- a/.github/docker/entrypoint.sh
+++ b/.github/docker/entrypoint.sh
@@ -56,9 +56,9 @@ if [ -z "$PAYMENTER_SKIP_DEFAULT" ] || [ "$PAYMENTER_SKIP_DEFAULT" != "true" ]; 
   echo -e "Renewing default themes and extensions..."
   
   # Renew default themes
-  if [ -d /app/themes_default ] && [ -d /app/themes ]; then
+  if [ -d /app/default/themes ] && [ -d /app/themes ]; then
     echo -e "Renewing themes from defaults..."
-    for item in /app/themes_default/*; do
+    for item in /app/default/themes/*; do
       if [ -e "$item" ]; then
         item_name=$(basename "$item")
         rm -rf "/app/themes/$item_name"
@@ -69,10 +69,10 @@ if [ -z "$PAYMENTER_SKIP_DEFAULT" ] || [ "$PAYMENTER_SKIP_DEFAULT" != "true" ]; 
   fi
   
   # Renew default extensions - only those that already exist
-  if [ -d /app/extensions_default ] && [ -d /app/extensions ]; then
+  if [ -d /app/default/extensions ] && [ -d /app/extensions ]; then
     echo -e "Renewing extensions from defaults..."
     updated_any=0
-    for item in /app/extensions_default/*; do
+    for item in /app/default/extensions/*; do
       if [ -e "$item" ]; then
         item_name=$(basename "$item")
         # Only update if extension category exists in /app/extensions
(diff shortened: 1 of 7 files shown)


BluetonicBeats:
closed since we both don't liked the approach

PR #1255, opened 2026-06-07
fix(pricing): prevent 500 when plan has no price in active currency

Adekabang:
Plan::price() dereferenced a null price row when a plan had no price in the requested currency (single-currency plans reached via stale cart/session or direct link), causing "read property setup_fee on null".

- Plan::price(): return an unpriced PriceClass instead of crashing.
- CartItem: resolve price against the cart's locked currency_code, not the session currency.
- Checkout: select the default plan from availablePlans() and auto-switch the session currency to one that prices the requested/first plan; redirect to the product page when none is available.


CorwinDev:
Superseded by 9a66ebf0a524004f12f4700389b73c37f9183757

PR #1261, opened 2026-06-09
Confirm payment on return from redirect checkout (fixes pending invoices without webhook)

### LINKED ISSUE #1260: [Bug]: Invoice stays "pending" after successful Stripe payment when webhook is unreachable
# Issue: Invoice stays "pending" after a successful Stripe payment (no webhook)

## Summary

After paying an invoice with Stripe, the invoice stays **pending** and the
service is never activated/provisioned unless the Stripe **webhook** reaches the
app. In local / test environments (where the webhook endpoint is usually not
reachable or not configured) a successful payment is effectively lost from the
app's point of view.

## Steps to reproduce

1. Set up Stripe in **test mode** on a local install (no public webhook URL, or
   webhook not forwarded via the Stripe CLI).
2. Create a pending invoice and pay it with a Stripe test card.
3. Stripe redirects back to `…/invoices/{id}?checkPayment=true&payment_intent=pi_…`.
4. The invoice remains **pending** indefinitely; the related service is not
   activated.

## Root cause

Marking an invoice paid is driven **only** by the
`payment_intent.succeeded` webhook (`Stripe::webhook()` →
`ExtensionHelper::addPayment()` → transaction created → invoice marked paid →
`ProcessPaidInvoiceService` activates the service).

The return page `app/Livewire/Invoices/Show.php` (`checkPaymentStatus()`) only
**polls** for `status === 'paid'`; it never confirms the payment with Stripe.
So with no webhook, nothing ever flips the invoice to paid.

## Expected behaviour

When the customer returns from Stripe's redirect checkout, the app should
confirm the payment directly with Stripe (the `payment_intent` is on the return
URL) and mark the invoice paid, without depending on the webhook.

## Proposed fix

Add a generic `Gateway::checkPayment(Invoice)` hook called on return from an
off-site checkout, and implement it for Stripe by fetching the PaymentIntent and
recording the payment. Idempotent (keyed by PaymentIntent id) so a later webhook
doesn't double-count. The webhook remains as the production source of truth /
fallback. See PR
([`stripe-pending-after-pay-pr.md`](./stripe-pending-after-pay-pr.md)).

## Impact

- Any deployment where the Stripe webhook is missing, misconfigured, or
  temporarily failing leaves paid invoices stuck pending and services
  un-provisioned.
- Especially affects local development and test/staging setups.

touhoku-japan:
# PR: Confirm payment on return from redirect checkout (Stripe)

Fixes #1260 — invoices getting stuck in **pending** after a successful Stripe
payment when the webhook doesn't reach the app.

> **Note:** stacked on top of #1259 (zero-decimal currency fix) because
> `Stripe::checkPayment()` reuses its `fromStripeAmount()` helper for the amount
> conversion. Merge #1259 first; this PR's diff then reduces to just the changes
> below.

## Problem

Invoices were only marked paid (and their services provisioned) by the Stripe
webhook (`payment_intent.succeeded`). The return page only polled for a paid
status, so when the webhook was unreachable/unconfigured — the normal case in
local and test environments — a successful payment left the invoice pending and
the service un-provisioned forever.

## Fix

- **`app/Classes/Extension/Gateway.php`** — add a generic, default no-op
  `checkPayment(Invoice $invoice): void` hook. Gateways that redirect off-site
  can override it to confirm a payment directly with the provider on return.
- **`app/Helpers/ExtensionHelper.php`** — add `checkPayment(Invoice $invoice)`
  that asks every gateway to verify the invoice (each gateway decides, from the
  request, whether it applies).
- **`app/Livewire/Invoices/Show.php`** — on return (`?checkPayment=true`) for a
  pending invoice, call `ExtensionHelper::checkPayment()` and refresh before
  falling back to polling.
- **`extensions/Gateways/Stripe/Stripe.php`** — implement `checkPayment()`:
  read the `payment_intent` Stripe appends to the return URL, fetch the intent,
  verify `metadata.invoice_id` matches, then record the payment
  (`succeeded → addPayment`, `processing → addProcessingPayment`).

## Why it's safe

- **Idempotent**: `addPayment` is keyed by the PaymentIntent id
  (`updateOrCreate` on `transaction_id`), so a webhook arriving later updates the
  same transaction instead of creating a duplicate — no double-counting.
- **No extra API calls on normal loads**: Stripe's `checkPayment` returns early
  unless `payment_intent` is present in the request (i.e. only on return from
  the redirect).
- **Webhook unchanged**: still the production source of truth and a fallback.
- **Ownership check**: only acts on an intent whose `metadata.invoice_id`
  matches the invoice being viewed.

## Testing

- `php -l` on all four changed files — no syntax errors.
- Manual (Stripe test mode, no webhook): pay a pending invoice, get redirected
  back, confirm the invoice flips to paid and the service is provisioned
  immediately.
- Regression: pay with the webhook configured and confirm no duplicate
  transaction is created (return-path + webhook resolve to the same
  PaymentIntent id).

### DIFF
--- a/app/Classes/Extension/Gateway.php
+++ b/app/Classes/Extension/Gateway.php
@@ -60,4 +60,18 @@ abstract class Gateway extends Extension
     {
         throw new \Exception('Not implemented');
     }
+
+    /**
+     * Verify a payment for the given invoice directly with the gateway.
+     *
+     * Called when the customer returns from an off-site / redirect checkout, so
+     * that activation does not depend on an asynchronous webhook (which may be
+     * unreachable in local / test environments). Implementations decide for
+     * themselves — based on the current request — whether they apply, and must
+     * be idempotent: recording the same payment again is safe. Default: no-op.
+     */
+    public function checkPayment(Invoice $invoice): void
+    {
+        // No-op by default; gateways that redirect off-site should override this.
+    }
 }
--- a/app/Helpers/ExtensionHelper.php
+++ b/app/Helpers/ExtensionHelper.php
@@ -406,6 +406,21 @@ class ExtensionHelper
         return self::getExtension('gateway', $gateway->extension, $gateway->settings)->charge($invoice, $invoice->remaining, $billingAgreement);
     }
 
+    /**
+     * Ask every gateway to verify whether the given invoice has been paid.
+     *
+     * Used when the customer returns from a redirect checkout to confirm the
+     * payment directly with the gateway, without waiting for a webhook. Each
+     * gateway decides for itself (based on the current request) whether it
+     * applies, so this is safe to call for any pending invoice.
+     */
+    public static function checkPayment(Invoice $invoice): void
+    {
+        foreach (Gateway::with('settings')->get() as $gateway) {
+            self::getExtension('gateway', $gateway->extension, $gateway->settings)->checkPayment($invoice);
+        }
+    }
+
     public static function getBillingAgreementGateways()
     {
         $gateways = [];
--- a/app/Livewire/Invoices/Show.php
+++ b/app/Livewire/Invoices/Show.php
@@ -36,7 +36,15 @@ class Show extends Component
     public function mount()
     {
         if (Request::has('checkPayment') && $this->invoice->status === 'pending') {
-            $this->checkPayment = true;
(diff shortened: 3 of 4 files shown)


CorwinDev:
Not planned

PR #1292, opened 2026-07-29
Notification Template Translations & Preferred Language

BluetonicBeats:
## Summary

Notification templates (email and in-app) support per-locale content in the admin UI. Users can store a preferred language on their account. That preference is used when resolving which template locale to send, including for queued and background notifications where no session is available.

## Data model

- Table `notification_template_translations` stores per-template, per-locale fields:
  - `subject`
  - `body`
  - `in_app_title`
  - `in_app_body`
  - `in_app_url`
  - `edit_preference_message`
- Columns on `notification_templates` hold the default-language content and act as fallback
- Column `users.preferred_language` stores the user’s preferred locale (nullable)

### Edit notification template

- Language dropdown (allowed languages only)
- Starts on the application default language
- Optional action: **Prefill from default language** (fills the form from default content; does not save until Save)
- For a non-default locale, empty subject and body on save remove that locale’s translation row (runtime falls back to default)

### Notification templates list

- Language badges show locales that have content
- Default language first, then other locales alphabetically
- At most five badges; additional locales are shown via a `[...]` badge with a tooltip
- All badges use the same style (uppercase locale codes)

## Locale resolution for notifications

When sending a notification:

1. Explicit `$data['locale']` if provided
2. `user.preferred_language`
3. Session `locale` when available
4. Application default language (`app_language` / `config('app.locale')`)

`NotificationTemplate::resolveForLocale()` then loads the matching translation, or falls back to the base template content.

## Preferred language

- Account page: preferred language select (allowed languages)
- Locale switcher (authenticated): updates session and `preferred_language`
- `SetLocale` middleware: session → preferred language → app default
- Admin user edit: preferred language field

## Known limitation

Guest locale changes are stored only in the session. Login does not copy `session('locale')` into `users.preferred_language`. The preference is written only when an authenticated user changes language (locale switcher or Account page).

| Situation | Result |
|-----------|--------|
| Same browser/session after login | UI may keep the guest language via session |
| `preferred_language` in the database | Unchanged |
| New session / other device / cleared cookies | Uses stored preference or app default, not the earlier guest choice |

### Visibility change

- `getAvailableLanguages()` was previously **`private`** and only used inside `Settings` for the Default Language / Allowed Languages settings UI
- It is now **`public static`** so other code (admin notification edit, Account page, User resource) can reuse the same `lang/*` directory scan without duplicating that logic

### New helpers

- `getAvailableLanguageOptions()` – maps installed locale codes to display labels from `config('app.available_locales')` (falls back to uppercase code)
- `getAllowedLanguageOptions()` – same mapping, restricted to `settings.allowed_languages` (intersected with installed locales)

These helpers feed language dropdowns and validation for allowed locales across admin and client UIs.

## Mail queue safety

`App\Mail\Mail` captures the resolved subject and body as plain strings before the message is queued, so queue workers do not reload untranslated base columns from the database.

### DIFF
--- a/app/Admin/Resources/NotificationTemplateResource.php
+++ b/app/Admin/Resources/NotificationTemplateResource.php
@@ -20,6 +20,7 @@ use Filament\Schemas\Schema;
 use Filament\Tables\Columns\IconColumn;
 use Filament\Tables\Columns\TextColumn;
 use Filament\Tables\Table;
+use Illuminate\Database\Eloquent\Builder;
 
 class NotificationTemplateResource extends Resource
 {
@@ -52,7 +53,7 @@ class NotificationTemplateResource extends Resource
                     ->collapsible()
                     ->schema([
                         TextInput::make('subject')
-                            ->required()
+                            ->required(fn ($livewire): bool => !($livewire instanceof EditNotificationTemplate) || $livewire->isEditingDefaultLocale())
                             ->maxLength(255),
                         Select::make('mail_enabled')
                             ->label('Mail Enabled')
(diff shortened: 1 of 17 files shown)


CorwinDev:
I think I'm gonna pass on this for now, most of the items aren't translatable (yet) (product names, category names etcetc)

BluetonicBeats:
> I think I'm gonna pass on this for now, most of the items aren't translatable (yet) (product names, category names etcetc)
> 
>  

That's right I was waiting for a while for a official option and build a 3rd party Tool which is way more extended then this approach but I still wanted to remind that this should be core functionality.

PR #1293, opened 2026-07-31
feat: Add required checkbox option for Config settings

romain-neil:
This change should allow admin to set an option to be required.

For example, this development allows customers to waive their right of withdrawal.

Please note that I will test this dev during this weekend, so do not hesitate to review the changes meantime


CorwinDev:
A config option checkbox is always optional, if you wish to add a "acceptance" button you'd need to use the terms of service url or code something yourself

PR #1295, opened 2026-08-02
Display credit balance in navbar/account page and fix credits_enabled enforcement

26bz:
Fixes `credits_enabled` still letting customers pay with credits at checkout even when the setting was unchecked, and fixes the balance not respecting the max credit cap. Also adds a navbar balance display and added/spent totals on the account credits page, and fixes the broken layout of the previous account credits page.

### DIFF
--- a/app/Classes/Settings.php
+++ b/app/Classes/Settings.php
@@ -490,11 +490,20 @@ class Settings
             ],
             'credits' => [
                 [
-                    'name' => 'credits_enabled',
-                    'label' => 'Credits Enabled',
+                    'name' => 'credits_payments_enabled',
+                    'label' => 'Credits enabled',
                     'type' => 'checkbox',
                     'database_type' => 'boolean',
                     'default' => false,
+                    'description' => 'Allow customers to pay with credits at checkout. Disabling this does not disable existing balances from being visible, but prevents them from being spent.',
+                ],
+                [
+                    'name' => 'credits_enabled',
+                    'label' => 'Credits topup enabled',
+                    'type' => 'checkbox',
+                    'database_type' => 'boolean',
+                    'default' => false,
+                    'description' => 'Allow customers to add (top up) credits to their account. Disabling this does not prevent existing balances from being spent.',
                 ],
                 [
                     'name' => 'credits_minimum_deposit',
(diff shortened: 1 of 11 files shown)


CorwinDev:
As we talked about in the discord, to prevent backwards compatibility make it so the current credits toggle is renamed to "Credits topup enabled", and then add a fresh checkbox to enable credit payments

26bz:
✅

app/Livewire/Services/Upgrade.php
```diff
@@ -209,13 +209,15 @@ public function doUpgrade()
 
                 $user = User::where('id', Auth::id())->lockForUpdate()->first();
                 $credit = $user->credits()->where('currency_code', $price->currency->code)->first();
+                $maxCredit = config('settings.credits_maximum_credit');
 
                 if ($credit) {
-                    $credit->increment('amount', abs($price->price));
+                    $credit->amount = min($credit->amount + abs($price->price), $maxCredit);
```
CorwinDev, line 215: I don't really like that we're not giving the client the amount of credits they need when they already have max credits

app/Services/Invoice/ProcessPaidInvoiceService.php
```diff
@@ -37,13 +37,15 @@ public function handle(Invoice $invoice): void
                 $user = $invoice->user;
                 $credit = $user->credits()->where('currency_code', $invoice->currency_code)->first();
 
+                $maxCredit = config('settings.credits_maximum_credit');
+
                 if ($credit) {
-                    $credit->amount += $item->price;
+                    $credit->amount = min($credit->amount + $item->price, $maxCredit);
```
CorwinDev, line 43: Same here

CorwinDev (changes requested):

app/Classes/Settings.php
```diff
@@ -489,12 +489,21 @@ public static function settings()
                 ],
             ],
             'credits' => [
+                [
+                    'name' => 'credits_payments_enabled',
+                    'label' => 'Credits enabled',
+                    'type' => 'checkbox',
+                    'database_type' => 'boolean',
+                    'default' => false,
```
CorwinDev, line 497: This should be true to retain the current behaviour

CorwinDev (changes requested):

PR #1302, opened 2026-08-08
Add 'expires_at' translation to services

Pexoa:
Add the missing 'expires_at' translation key to services.php for multiple locales (ar, bn, da, de, en, es, fi, fr, he, hi, hu, id, it, ko, lv, nl, no, pl, pt, sr, sv, tr, uk, zh) so UIs can display expiration dates in each language.

### DIFF
--- a/lang/ar/services.php
+++ b/lang/ar/services.php
@@ -57,6 +57,7 @@ return [
     'price_one_time' => ':price مرة واحدة',
     'renews_in' => 'Renews in',
     'renews_on' => 'Renews on',
+    'expires_at' => 'تنتهي في',
     'auto_pay' => 'دفع تلقائي بإستخدام',
     'auto_pay_not_configured' => 'لم يتم الإعداد',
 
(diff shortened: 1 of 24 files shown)


CorwinDev:
`expires_at` key doesn't seem to be used (anymore)?

Pexoa:
No, it’s not being used anymore, as you can see in the image the `expires_at` field is still being displayed under Product Details, but there doesn’t seem to be any actual usage of it.

image

CorwinDev:
That is a custom theme which we don't support in core code, please ask the author to either hardcode it or include translation files

Pexoa:
I did speak to Buzz (the theme author) about this. He confirmed that his theme doesn't define `services.expires_at` itself it's referencing a translation key that should exist in Paymenter core's `lang` folder but doesn't. The default theme works fine only because it doesn't use that key at all. The moment you enable Buzz's theme, it breaks because the key is missing from core.

Buzz told me to add the translation myself, which is exactly why I've raised this PR — the key is missing from Paymenter core and should be added there.

CorwinDev:
We've removed this key in a previous version because its unused in the default theme, there are no plans to bring it back

Pexoa:
No problem, I'll sort it myself directly in the theme file. Thanks for clarifying @CorwinDev!

PR #1305, opened 2026-08-16
fix: resolve cart item prices against the cart's locked currency

gabbelitoV2:
## Summary

Cart line amounts are resolved from `session('currency')` at checkout time, while the order/invoice/service are created with the **cart's** `currency_code` (stamped once at cart creation — and the cart cookie lives 30 days, far longer than the session). When the session expires and falls back to `settings.default_currency`, a returning visitor with a surviving cart gets invoiced **default-currency amounts labeled with the cart's currency**.

This is not theoretical — it hit a production store last week: a customer added a plan priced at **3.99 EUR / 39.00 SEK** to a SEK-stamped cart, came back the next day after his session had expired, and checked out. The invoice (and the gateway charge) came out as **3.99 SEK** — the EUR amount with the SEK label, a 90% accidental discount. No coupon involved; every guard in `CheckoutParameterMiddleware`/`LocaleSwitch` only blocks *active* currency switches while the cart has items, so none of them catches the silent session-expiry fallback.

## What this PR does

- **`CartItem::price()`** resolves the plan price **and** the config-option child prices against `$this->cart->currency_code` (session only as fallback when the item has no cart). The invoice inherits the cart's currency, so the amounts must come from it too.
- **`Plan::price()`** returns an unavailable `Price` (currency `null` → `available === false`, renders "Not available in your currency") instead of dereferencing `null` when no price row exists for the resolved currency.
- **`HasPlans::price()`** filters its plan lookups by the same currency it prices in. Previously, when a `currency:` argument was passed but the billing-period lookup (which used session currency) found nothing, the fallback loop silently priced a **different billing period** (repro: a yearly config option priced at the monthly rate). It now reports unavailable instead.
- **`Cart::checkout()`** refuses items whose price is unavailable in the cart's currency. Previously such items resolved to price `0` and checked out **free** while still being provisioned.
- New `lang/en/product.php` key `not_available` for the checkout error.

## Tests

`tests/Feature/CartCurrencyTest.php` — two regression tests, both **fail on current master and pass with this fix**:

1. `test_item_price_follows_cart_currency_when_session_diverges` — cart stamped USD, session flipped to EUR: the item must price at the USD row with USD currency.
2. `test_item_without_price_row_in_cart_currency_is_unavailable` — cart locked to a currency the plan has no row for: the item must resolve `available === false` instead of silently pricing 0.

Full suite run before/after: the only delta is the two new tests going red → green (the handful of pre-existing failures in my environment are `ViteManifestNotFoundException` render tests, identical on clean master).

The `CartItem`/`Plan` changes have also been running patched on a production v1.5.7 store, verified against its live data (multi-currency EUR/SEK with priced config options).

## Relation to earlier work

- #1255 proposed resolving cart prices against the cart's locked currency and was closed as superseded by 9a66ebf — but 9a66ebf only adds the availability redirect on the **product page** (`Products/Checkout::mount`); the cart/checkout path it doesn't touch is where the invoice amounts are written, and the incident above happened on a version that includes 9a66ebf.
- #1277 reported the 500 when no price row exists in the session currency; the `Plan::price()` null-guard here fixes that class of crash as well.

### DIFF
--- a/app/Livewire/Cart.php
+++ b/app/Livewire/Cart.php
@@ -126,6 +126,11 @@ class Cart extends Component
             }
             // Lock the orderproducts
             foreach ($cart->items as $item) {
+                // An item without a price row in the cart's currency would otherwise check out free.
+                if (!$item->price->available) {
+                    throw new DisplayException(__('product.not_available', ['product' => $item->product->name]));
+                }
+
                 // Make sure we have the latest product data and lock it
                 $product = Product::where('id', $item->product->id)->lockForUpdate()->first();
 
--- a/app/Models/CartItem.php
+++ b/app/Models/CartItem.php
@@ -46,17 +46,24 @@ class CartItem extends Model
     {
         return Attribute::make(
             get: function () {
+                // Resolve against the cart's currency: the invoice inherits it, while the session may have expired back to the default currency.
+                $currency = $this->cart?->currency_code ?? session('currency', config('settings.default_currency'));
                 $total = 0;
                 $setup_fee = 0;
-                $total += $this->plan->price()->price;
-                $setup_fee += $this->plan->price()->setup_fee;
-                $this->product->configOptions->each(function ($option) use (&$total, &$setup_fee) {
+                $unavailable = false;
+                $total += $this->plan->price($currency)->price;
+                $setup_fee += $this->plan->price($currency)->setup_fee;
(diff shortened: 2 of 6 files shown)


CorwinDev:
Can you reduce the amount of comments? One line on hard to understand code is more then enough

gabbelitoV2:
Done trimmed everything down to single-line comments.

PR #1312, opened 2026-08-20
Add configurable automatic renewal option per plan

WallyTheBot:
This pull request introduces a configurable auto_renew toggle per plan in the admin panel and integrates it into the automated cron job processing.

Example use:
For example, if you set a plan to $0, it’s essentially free, and you can also choose the lifetime option. If you want to go with a monthly plan, it’s essentially the same as the lifetime plan. And if you go with a monthly plan, you can encourage users to actively renew the plan themselves—essentially as a free plan—but they have to take the initiative themselves.

Changes
- Database Migration: Added the `auto_renew` boolean column to the `plans` table (defaulting to `true`).
- Admin Panel Resource: Updated `ProductResource` (within the plan repeater schema) to allow administrators to enable or disable automatic renewals individually for each plan.
- Model Update: Added `auto_renew` to the `$fillable` attributes in the `Plan` model.
- Cron Job Logic: Adjusted the `CronJob` console command so that when `auto_renew` is disabled (`false`) for a specific plan, renewal invoices are still generated, but automatic payments (via credits and payment gateways/billing agreements) are skipped.

### DIFF
--- a/app/Admin/Resources/ProductResource.php
+++ b/app/Admin/Resources/ProductResource.php
@@ -20,6 +20,7 @@ use Filament\Forms\Components\RichEditor;
 use Filament\Forms\Components\Select;
 use Filament\Forms\Components\Textarea;
 use Filament\Forms\Components\TextInput;
+use Filament\Forms\Components\Toggle;
 use Filament\Infolists\Components\TextEntry;
 use Filament\Notifications\Notification;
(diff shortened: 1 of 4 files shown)


CorwinDev:
That doesn't really make sense? How are they gonna renew without a button or anything

PR #1322, opened 2026-09-05
fix: use unambiguous date formats for cancellation notice and invoice PDFs

fastcat-servers:
Making Paymenter a little more friendly for Americans by adopting translated date format in missing places (admin service cancellation pending notice and invoice PDFs).

**Before:**

**After:**


CorwinDev:
Huge QOL improvement for our freedom friends

PR #1323, opened 2026-09-07
fix(pdf): display specific service label on invoice items

WallyTheBot:
### Summary
This pull request updates the PDF invoice template to display the specific service label (retrieved via the polymorphic `reference` relation) beneath the standard item description.

### Changes
- Updated `resources/views/pdf/invoice.blade.php` to check for `$item->reference->label`.
- Added a sub-label display under the main invoice item description.

### Motivation
Previously, invoices only showed the generic item/product name (e.g., "Python"), making it difficult to distinguish between multiple instances of the same service. This change ensures users can clearly see the exact service label on their invoices.


resources/views/pdf/invoice.blade.php
```diff
@@ -153,7 +153,14 @@
         <tbody>
             @foreach($invoice->items as $item)
             <tr>
-                <td>{{ $item->description }}</td>
+                <td>
+                    {{ $item->description }}
+                    
+                    {{-- Holt das Label direkt über die polymorphe Service-/Reference-Relation --}}
+                    @if($item->reference && !empty($item->reference->label))
+                        <br><small style="color: #666; font-size: 0.9em;">{{ $item->reference->label }}</small>
```
CorwinDev, line 160: Label is always set, either to the customers configured label or the Product name label. 

So when the customer didn't configure any label it will display it duplicate

CorwinDev (changes requested):

CorwinDev:
Well, item description contains the dates too, so it will still always show :D

CorwinDev:
Yup that makes sense, lets remove the redudant check for same description

CorwinDev (approved):

PR #1326, opened 2026-09-13
Refresh the config options of a cart item whose product changed

### LINKED ISSUE #1207: [Bug]: If config options changes when users already have a item in cart, it will give errors like "The selected .... is invalid".
### Checklist

- [x] I have updated to the latest available Paymenter version.
- [x] I have cleared the cache of my browser.
- [x] I have tried a different browser to see if it is related to my browser.

### Describe the problem you are having

 If config options changes when users already have a item in cart, it will give errors like "The selected .... is invalid". I fixed this by manually adding a auto cart refresh on config option change. So it will always use the updated config options, please do this.

### Steps to reproduce

1. Add an item to cart
2. Change config options
3. Refresh cart page and try checkout
...

### Version

1.4.7

### In which browser(s) are you experiencing the issue with?

_No response_

### Relevant Paymenter log output

```shell
No log from paymenter.
```

### Operating system

Linux

### Any other information that may be helpful

_No response_

dualfroz:
Reopening a cart item kept its stored config values even when the product no longer offered them, so checkout failed with "The selected ... is invalid".
Stored values are now kept only while still offered, otherwise the first available one is used. Checking out straight from the cart page is not covered.

closes #1207


CorwinDev:
I don't know if its a good idea to force the first instead of letting the customer choose an alternative?

dualfroz:
it's the same default a new item gets, and the customer can still change it before checkout. can make it force a pick instead, but the select would need an empty option in the theme

PR #1329, opened 2026-09-17
feat(tickets): add custom service label display in ticket service dro…

WallyTheBot:
### Summary
This PR adds support for custom service labels (`$service->label`) across all ticket service selection dropdowns and infolists

### Changes
- **Admin Panel (`app/Admin/Resources/TicketResource/Pages/EditTicket.php`):**
  - Updated both the `infolist` service entry and the edit action modal select field to utilize the unified `Product Name (Label) (Status)` nomenclature using `getOptionLabelFromRecordUsing`.

### Benefits
For hosting providers offering multiple instances of identical plans (e.g., multiple game servers or web hosting spaces), clients and administrators can now instantly distinguish between them using their user-defined custom labels.

### i need help
** I was also planning to customize the dropdown in `admin/tickets/create`, but there isn't much in the file `app/Admin/Resources/TicketResource/Pages/CreateTicket.php`. Could you tell me where I need to add it so that it appears on all ticket pages? I'd then include that in this pull request.**


CorwinDev:
It doesn't really make sense to use the customer defined label on the admin side? It might be a good idea to add the id but the label is kinda wrong

